PR #106, opened 2019-02-17
Frontend Stylistic Fixes

mt-xing:
- Overflow hidden on the page, removed shadow behind progress tracker
- Add class box clears after adding class


mt-xing:
It seems that while deleting the tag selected in add task no longer crashes, any active tasks tagged with that tag will still cause a crash. To the reviewer: don't let me merge this branch without fixing that.

SamChou19815:
> It seems that while deleting the tag selected in add task no longer crashes, any active tasks tagged with that tag will still cause a crash. To the reviewer: don't let me merge this branch without fixing that.

@mt-xing I think it's a regression. It happens before when I rewrite the backend I forget to also modify all tasks with that tag. Do you want me to fix in my branch?

mt-xing:
@SamChou19815 If you get a chance, that would be ideal.

SamChou19815:
> @SamChou19815 If you get a chance, that would be ideal.

I will likely to push the changes in minutes. After that, you can review my pull request and merge the changes to your branch.

mt-xing:
So due to my flow type checker breaking after the latest merge, I've been unable to get the tag fix working within this branch by tonight. Instead, we can just merge the changes I already have as is.

PR #147, opened 2019-03-05
made pin in future view light

meganyin13:


SamChou19815:
You can merge it if you think it's ready

PR #148, opened 2019-03-05
made settings button 17.5px

meganyin13:
as per designers' spec

### DIFF
--- a/frontend/src/assets/svgs/settings.svg
+++ b/frontend/src/assets/svgs/settings.svg
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<svg width="200px" height="200px" viewBox="0 0 36 36" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+<svg width="36px" height="36px" viewBox="0 0 36 36" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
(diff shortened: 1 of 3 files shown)


frontend/src/assets/svgs/settings.svg
```diff
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<svg width="36px" height="36px" viewBox="0 0 36 36" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+<svg width="200px" height="200px" viewBox="0 0 36 36" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
SamChou19815: You don't need to change the svg dimension. One webpack plugin will automatically strip those dimensions away from svg component. CSS modification should be sufficient.

meganyin13: I realized that as I was testing it and added the CSS modification to SettingsButton.css

frontend/src/components/TitleBar/Settings/SettingsButton.jsx
```diff
@@ -15,7 +15,7 @@ export default function SettingsButton() {
   return (
     <div style={{ display: 'inline-block' }}>
       <button type="submit" onClick={displayModal}>
-        <Settings />
+        <Settings className={styles.SettingsButton} />
```
SamChou19815: @meganyin13 I changed this line. The css module names are case-sensitive. I think this pull request is ready to go.

PR #155, opened 2019-03-09
Create admin control panel

SamChou19815:
This pull request creates an admin control panel. You are able to run it locally with a service account. Although the deploy script is already written, deployment cannot be done right now due to a missing billing account.

Current UI:
Screen Shot 2019-03-09 at 4 31 37 PM


SamChou19815:
This branch is already for review and merging

Issue #96, opened 2019-02-13
Open Source License


SamChou19815:
We should consider adding an open source license.

PR #156, opened 2019-03-09
Create MIT License. Fix (#96)

SamChou19815:
Only accept this pull request if we decide to adopt the MIT license.


SamChou19815:
MIT license is no longer considered after today's TPM meeting.

PR #215, opened 2019-04-01
added 3 analytics functions

jessicahong9:
certain fields will need to be added to the tasks in the database in order for functions to work. For example, "dateCompleted" needs to be added in order to return "time saved" by the user.


SamChou19815:
@jessicahong9 you have accidentally committed the `backend` submodule. Could you remove it?

jessicahong9:
should be good now!

PR #216, opened 2019-04-09
working on settingsbutton...

meganyin13:


frontend/src/components/UI/Tooltip.tsx
```diff
@@ -0,0 +1,56 @@
...
+  public constructor(props: any) {
+    super(props);
+    this.state = {
+      name: props.name,
+      iconName: props.iconName,
+      hover: false,
+    };
+  }
+
+  public handleMouseIn() {
```
SamChou19815: You can write with arrow function like
```typescript
private handleMouseIn = () => { /* code */ }
```
so you don't need to use bind

frontend/src/components/UI/Tooltip.tsx
```diff
@@ -0,0 +1,56 @@
...
+  iconName: IconName;
+  hover: boolean;
+}
+
+class Tooltip extends React.Component<props, states> {
+  public constructor(props: any) {
+    super(props);
+    this.state = {
+      name: props.name,
+      iconName: props.iconName,
```
SamChou19815: You are duplicating props' field into state, which can easily go out of sync. Do you have compelling reasons to do that?

frontend/src/components/UI/Tooltip.tsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import SamwiseIcon from './SamwiseIcon';
+import { IconName } from './Samwise-icon-types';
+
+type props = {
```
SamChou19815: Please use `type Props` to keep the codebase consistent

PR #226, opened 2019-04-16
Made the Favicon Square

mt-xing:
Let me double check that everything works before merging


mt-xing:
Should be good to go. 256x256 max size favicon.

PR #236, opened 2019-04-21
[WIP] Analysis Scripts

jessicahong9:


admin-scripts/addLastEdited.py
```diff
@@ -0,0 +1,19 @@
...
+from firebase_admin import db
+import datetime
+from firebase_admin import firestore
+
+cred = credentials.Certificate('firebaseadmin.json')
+firebase_admin.initialize_app(cred)
+
+db = firestore.client()
+tasks = db.collection(u'samwise-tasks').get()
+
```
SamChou19815: @jessicahong9 could you change the code so that it uses `serverTimestamp`? Also, could you also add the new field for `tags` and `settings`.

PR #241, opened 2019-04-22
Types for Repeating Task

meganyin13:


frontend/src/store/store-types.ts
```diff
@@ -30,8 +30,39 @@ export type Task = {
...
+export type PartialTask = Partial<Pick<Task, MainTaskProperties | 'children' | 'type' >>;
+
+export type RepeatMetaData = {
+  readonly startDate: Date;
+  readonly endDate: Date;
+  readonly pattern: RepeatingPattern;
+}
+
+export type ForkedTask = PartialTask & {
+  readonly masterID: string;
```
SamChou19815: No need for master ID here, because the id is always there in the Task type

frontend/src/store/store-types.ts
```diff
@@ -30,8 +30,39 @@ export type Task = {
...
+/**
+ * The task type without id and every field optional.
+ * Fields that are filled represent differences from master template
+ */
+
+export type PartialTask = Partial<Pick<Task, MainTaskProperties | 'children' | 'type' >>;
+
+export type RepeatMetaData = {
+  readonly startDate: Date;
+  readonly endDate: Date;
```
SamChou19815: endDate can potentially be `null`

frontend/src/store/store-types.ts
```diff
@@ -30,8 +30,39 @@ export type Task = {
   readonly complete: boolean;
   readonly inFocus: boolean; // Whether the task is in focus
   readonly children: Set<string>;
+  readonly type: 'ONE_TIME' | null; // Null for legacy
```
SamChou19815: You should not put the type here. Similar to the TypeScript in the wiki, you should extract a common Task type and then have something like
```typescript
type RepeatedTask = CommonTask & {
  readonly type: 'REPEATED';
  // ...
}
type OneTimeTask = CommonTask & {
  readonly type: 'ONE_TIME';
  // ...
}
type LegacyTask = CommonTask;
type Task = RepeatedTask | OneTimeTask | LegacyTask;
```

frontend/src/store/store-types.ts
```diff
@@ -30,8 +30,39 @@ export type Task = {
...
+
+export type PartialTask = Partial<Pick<Task, MainTaskProperties | 'children' | 'type' >>;
+
+export type RepeatMetaData = {
+  readonly startDate: Date;
+  readonly endDate: Date;
+  readonly pattern: RepeatingPattern;
+}
+
+export type ForkedTask = PartialTask & {
```
SamChou19815: Probably name it `ForkedTaskMetadata` because it does not contain all the info of the forked task

frontend/src/store/store-types.ts
```diff
@@ -73,7 +108,7 @@ export type Course = {
  */
 export type State = {
   readonly tags: Map<string, Tag>;
-  readonly tasks: Map<string, Task>;
+  readonly tasks: Map<string, CommonTask>;
```
SamChou19815: It should be the union type of `LegacyTask`, `OneTimeTask` and `RepeatedTask`

SamChou19815: Other than that, the types look good to me.

PR #245, opened 2019-04-24
TaskEditor v2 with partial support for repeated task

SamChou19815:
- [x] create the reducer for task editor while tracking diff
- [x] connect the reducer to the task editor with `React.useReducer`.
- [x] pass in data with complete information about the type of the task.
- [x] prompt the user to decide whether to apply change to master or save locally if necessary.
- [x] connect to Megan's firebase actions to ensure the save is applied. Call different functions based on user response.

### DIFF
--- a/frontend/.eslintrc
+++ b/frontend/.eslintrc
@@ -30,7 +30,10 @@
   },
   "settings": {
     "import/resolver": {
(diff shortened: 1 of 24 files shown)


SamChou19815:
Expect to see a lot of type errors since we are still in the process of refactoring the types to support repeated tasks

PR #247, opened 2019-04-24
Firebase actions

meganyin13:


frontend/src/firebase/actions.ts
```diff
@@ -152,21 +139,51 @@ export const removeAllForks = (taskId: string): void => {
...
+      await batch.commit();
+      const subtaskIds = createdSubTasks.map(s => s.id);
+      const taskWithChildren = { ...repeatingTask, children: subtaskIds };
+      const firestoreTask: FirestoreTask = await createFirestoreObject('tasks', taskWithChildren);
+      const addedDoc = await tasksCollection().add(firestoreTask);
+      const newForkMetaData = {
+        forkId: addedDoc.id,
+        replaceDate: mainTaskEdits.date !== undefined ? mainTaskEdits.date : repeatingTask.date,
+      };
+      batch.update(tasksCollection().doc(taskId), {
```
SamChou19815: You cannot use batch again. The `batch` is already committed above. At this point, you should think of this as an already closed file and you cannot do further operations on it.

frontend/src/firebase/actions.ts
```diff
@@ -152,21 +139,51 @@ export const removeAllForks = (taskId: string): void => {
...
   taskId: string,
-  forMasterTemplate: boolean,
+  editType: EditType,
   { mainTaskEdits, subTaskCreations, subTaskEdits, subTaskDeletions }: Diff,
 ): void => {
   const batch = db().batch();
-  if (forMasterTemplate) {
+  let updateTaskId = taskId;
+  if (editType === 'FORKING_MASTER_TEMPLATE') {
+    const { tasks, subTasks } = store.getState();
```
SamChou19815: I personally think it might be more efficient to apply the diff to a task fetch from the store to create a complete forked task, then call `addTask` above to actually add it. It has the added benefit of reusing the undo logic

frontend/src/firebase/actions.ts
```diff
@@ -152,21 +139,51 @@ export const removeAllForks = (taskId: string): void => {
...
+      };
+      batch.update(tasksCollection().doc(taskId), {
+        forks: [...repeatingTask.forks, newForkMetaData],
+      });
+      updateTaskId = addedDoc.id;
+    })();
+  }
+
+  if (editType === 'EDITING_MASTER_TEMPLATE') {
     removeAllForks(taskId);
```
SamChou19815: You will enter a race condition below. If you look at the implementation of `removeAllForks`, it is not a blocking function. I think the easiest way is to change `removeAllForks` into an async function, and then `await` for `await removeAllForks(taskId)` (although you might need to make the entire block starting here to be `async`).

frontend/src/firebase/actions.ts
```diff
@@ -182,6 +170,39 @@ export const editTaskWithDiff = (
...
+  { mainTaskEdits, subTaskCreations, subTaskEdits, subTaskDeletions }: Diff,
+): void => {
+  const batch = db().batch();
+  if (editType === 'FORKING_MASTER_TEMPLATE') {
+    (async () => {
+      const { tasks } = store.getState();
+      const repeatingTaskMaster = tasks.get(taskId) as RepeatingTask;
+      const { id, ...taskWithoutID } = repeatingTaskMaster;
+      const newTaskId = getNewTaskId();
+      batch.set(tasksCollection().doc(newTaskId), { newTaskId, ...taskWithoutID }, { merge: true });
```
SamChou19815: this spread is not safe. `...taskWithoutID` also contains `type: 'MASTER_TEMPLATE', repeats: ...`. You need to exclude them and then add `type: 'ONE_TIME'` for the new task.
It's advised that you write something like 
```typescript
const newOneTimeTask: FirestoreOneTimeTask = {...}
```
to ensure you indeed have all the required stuff and nothing more.

SamChou19815: Other than that, things are looking good.

frontend/src/firebase/actions.ts
```diff
@@ -182,6 +171,51 @@ export const editTaskWithDiff = (
...
+      const repeatingTaskMaster = tasks.get(taskId) as RepeatingTask;
+      const { id, type, children, ...taskWithoutId } = repeatingTaskMaster;
+      const createdSubTasks: Set<string> = children.map((subTaskId) => {
+        const subTask = subTasks.get(subTaskId);
+        const newSubTaskDoc = subTasksCollection().doc(subTaskId);
+        const firebaseSubTask: FirestoreSubTask = mergeWithOwner(subTask);
+        batch.set(newSubTaskDoc, firebaseSubTask);
+        return newSubTaskDoc.id;
+      });
+      await batch.commit();
```
SamChou19815: You should remove this line because you are committing the same `batch` below

PR #259, opened 2019-05-04
Finished Task Creator for Repeating Tasks

mt-xing:
Finished the basic task creator for creating repeating tasks. Includes fixes for the Date Picker component, so that it computes end dates correctly. Does not have a way of storing how the user wanted the repeat to end yet (so it defaults to end date, rather than storing the number of occurrences, for example). Also right now it's number of weeks to repeat, not number of occurrences (to simplify math).

### Breaking Changes

- Fixed a bug with the bitset utils that were off by one with day indexing. However, this fix seems to have caused problems in the future view, which was evidently expected the broken version of the bitset utils.

### Checklist:

- [X] My code follows the code style of this project.
- [ ] My changes requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] I tested affected functionality.
- [X] I resolved any merge conflicts.
- [X] My PR is ready for review.

### DIFF
--- a/frontend/src/util/bitwise-util.ts
+++ b/frontend/src/util/bitwise-util.ts
@@ -6,15 +6,15 @@
 export const DAYS_IN_WEEK = 7;
 
 export const isBitSet = (bit: number, d: number, totalLen: number): boolean => (
-  (bit & (1 << (totalLen - d))) !== 0
+  (bit & (1 << (totalLen - 1 - d))) !== 0
 );
 
 export const setBit = (bit: number, index: number, totalLen: number): number => (
-  bit | (1 << (totalLen - index))
+  bit | (1 << (totalLen - 1 - index))
 );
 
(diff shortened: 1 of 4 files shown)


frontend/src/util/bitwise-util.ts
```diff
@@ -6,15 +6,15 @@
 export const DAYS_IN_WEEK = 7;
 
 export const isBitSet = (bit: number, d: number, totalLen: number): boolean => (
-  (bit & (1 << (totalLen - d))) !== 0
+  (bit & (1 << (totalLen - 1 - d))) !== 0
```
SamChou19815: Could you write some explanation for these bitwise operations and then write some unit tests for it?

mt-xing: Test cases have been added.

PR #289, opened 2019-09-22
Clean Up Date Picker Repeating Task Code

mt-xing:
Cleaned up a lot of the React state hooks inside the date picker component. Consolidated into one state type that more closely mirrors what gets stored to the database.

### Inside This PR

- [X] Replaced individual state hooks with hook tracking one internal date type.
- [X] Removed a lot of unnecessary fragment wrapping

### Notes 

The type I created stores all info for both types of tasks at all times, rather than a union type, so that if the user clicks from one type to another and then comes back, what they previously entered is not lost. A similar reasoning is used to justify why repeatEnd always stores both the end date and the number of weeks. This type is local only to this component; extraneous data is thrown out before the database write.

### Checklist:

- [X] My code follows the code style of this project.
- [X] I tested affected functionality.
- [X] I resolved any merge conflicts.
- [x] My PR is ready for review.


frontend/src/components/TaskCreator/DatePicker.tsx
```diff
@@ -65,142 +100,130 @@ export default function DatePicker(props: Props): ReactElement {
...
-
   /**
    * Event handler for when the user starts the repeat box
    */
   const changeRepeat = (e: ChangeEvent): void => {
-    setIsRepeat((e.target as HTMLSelectElement).value === 'true');
+    setInternalDate({
+      ...internalDate,
+      type:
+       (e.target as HTMLSelectElement).value === 'true'
```
SamChou19815: There are definitely better way than this cast.
Usually I use `e.currentTarget.value` for this. `ChangeEvent` type can be rewritten as `ChangeEvent< HTMLSelectElement>`.

frontend/src/components/TaskCreator/DatePicker.tsx
```diff
@@ -280,67 +316,55 @@ export default function DatePicker(props: Props): ReactElement {
...
-        endDate,
-        pattern: { type: 'WEEKLY', bitSet },
-      };
-      onDateChange(repData);
+    if (internalDate.repeatEnd.type === 'date') {
+      endDate = internalDate.repeatEnd.date;
     } else {
-      onDateChange(currDate);
+      // TODO once database support exists, replace this with number of occurrances
+      endDate = new Date(+(new Date()) + 1000 * 60 * 60 * 24 * 7 * internalDate.repeatEnd.weeks);
```
SamChou19815: Is this `+` JS type coercion magic?

SamChou19815:
@mt-xing is this ready?

mt-xing:
I think so

PR #293, opened 2019-09-27
[WIP] Get rid of save button

meganyin13:
### Summary 
Got rid of save button and extra space formatting for save. Added save functionality to onMouseLeave.

This pull request is the first step towards implementing feature Foo

- [x] implemented X
- [ ] fixed Y

### Breaking Changes  

- Database schema change (anything that changes Firestore collection structure)
- Other change that could cause problems (Detailed in notes)

### DIFF
--- a/frontend/src/components/Util/TaskEditors/TaskEditor/index.tsx
+++ b/frontend/src/components/Util/TaskEditors/TaskEditor/index.tsx
@@ -8,7 +8,6 @@ import { connect } from 'react-redux';
 import { MainTask, State, SubTask, Tag } from 'store/store-types';
 import OverdueAlert from 'components/UI/OverdueAlert';
 import { NONE_TAG } from 'util/tag-util';
-import { ignore } from 'util/general-util';
 import { confirmRepeatedTaskEditMaster, promptRepeatedTaskEditChoice } from 'util/task-util';
 import { editTaskWithDiff, forkTaskWithDiff } from 'firebase/actions';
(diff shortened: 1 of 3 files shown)


frontend/src/components/Util/TaskEditors/TaskEditor/index.tsx
```diff
@@ -94,11 +94,11 @@ function TaskEditor(
...
-  const onMouseLeave = (): void => {
-    if (onBlur) {
-      onBlur();
-    }
-  };
+  // const onMouseLeave = (): void => {
+  //   if (onBlur) {
+  //     onBlur();
+  //   }
+  // };
```
SamChou19815: Do not leave commented code. We have version control.

mt-xing:
Does tabbing out of the dialog box work as expected?

meganyin13:
@mt-xing I have pushed changes to add `onSaveClicked` to `onBlur` but when adding a new subtask when you type one letter it jumps out, saves that, and then you have to click in to finish editing. Changes are saved though when tabbing out.

mt-xing:
It's a little janky, what with needing to move the mouse away before it actually saves, rather than just checking off the task when clicked as a user might expect. We can talk about this more tomorrow. Good work so far.

SamChou19815:
Looks like more work needs to be done on the `TaskEditor` component. I will still lose focus when I press enter in the last subtask input box.

PR #297, opened 2019-09-29
Setup CD workflow with GitHub Actions and remove circle ci config

SamChou19815:
### Summary 

- [x] Setup GitHub Actions CD workflow
- [x] Remove Circle CI config to reduce gh-pages branch noise.
- [x] Replaced Circle CI badge with GitHub Actions badge

### Test Plan 

- For the deployment workflow, we can only test in `master` 🤔 
- See badge docs https://help.github.com/en/articles/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository


SamChou19815:
@mt-xing Circle CI is always failing on gh-pages branch because there is no circle ci config there and it's super annoying to receive those build failed emails. It also does not make a lot of sense to keep two CI system that is doing the same thing.
My guess is circle CI's cache might be broken. GitHub Actions work fine since it's guaranteed to run from a clean environment.

PR #300, opened 2019-10-02
Use firebase analytics

SamChou19815:
### Summary 

Firebase finally released analytics for web. Let's do everything in Firebase.
This makes ReactGA unused, so I removed it from package.json.

### Test Plan 

Analytics still works:
Screen Shot 2019-10-02 at 17 36 28

### DIFF
--- a/frontend/src/firebase/index.ts
+++ b/frontend/src/firebase/index.ts
@@ -10,6 +10,8 @@ if (process.env.NODE_ENV === 'production' && process.env.REACT_APP_IS_STAGING !=
     projectId: 'dti-samwise',
     storageBucket: 'dti-samwise.appspot.com',
     messagingSenderId: '114434220691',
+    appId: '1:114434220691:web:01579d764b20fb0aeb6366',
+    measurementId: 'G-DHQY39TSMH',
   };
 } else {
   firebaseConfig = {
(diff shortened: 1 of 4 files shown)


frontend/src/util/ga-util.ts
```diff
@@ -1,31 +1,44 @@
...
-    gaOptions: {
-      sampleRate: 100,
-    },
-  });
-  ReactGA.pageview(window.location.pathname + window.location.search);
+  firebase.analytics().logEvent('screen-view', { screenName: 'Homepage' });
 };
 
-export const setGAUser = ({ uid }: AppUser): void => ReactGA.set({ userId: uid });
+type EventType =
```
SamChou19815: This is used for custom event. If I don't define it here and put it in type parameter in line 28, typescript will not be happy.

frontend/src/firebase/index.ts
```diff
@@ -10,6 +10,8 @@ if (process.env.NODE_ENV === 'production' && process.env.REACT_APP_IS_STAGING !=
     projectId: 'dti-samwise',
     storageBucket: 'dti-samwise.appspot.com',
     messagingSenderId: '114434220691',
+    appId: '1:114434220691:web:01579d764b20fb0aeb6366',
```
mt-xing: Just double checking that you intended to expose these.

SamChou19815: Yes, this is specified in the analytics docs: https://firebase.google.com/docs/analytics/get-started

PR #303, opened 2019-10-05
Better and more snapshot tests

SamChou19815:
### Summary

- [x] Add new redux store with some values inside just for testing
- [x] Modify existing tests to use the testing redux store.
- [x] Add more snapshot tests in `TitleBar/Tags`

### Test Plan

The added tests are all passing! :)

I'm originally planning to add more tests on the big `FutureView` component, but it will be very flaky since it will render different stuff depending on the current time. To keep the diff small, I'm planning to leave it to the future.

### DIFF
--- a/frontend/src/components/TitleBar/Settings/SettingsPage.test.tsx
+++ b/frontend/src/components/TitleBar/Settings/SettingsPage.test.tsx
@@ -1,26 +1,23 @@
 import React from 'react';
 import { Map } from 'immutable';
-import { Provider } from 'react-redux';
 import renderer from 'react-test-renderer';
 import { Tag } from 'store/store-types';
-import { store } from 'store/store';
+import { ProviderForTesting } from 'store';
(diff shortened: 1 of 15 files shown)


frontend/src/components/TitleBar/Settings/SettingsPage.test.tsx
```diff
@@ -1,26 +1,23 @@
 import React from 'react';
 import { Map } from 'immutable';
-import { Provider } from 'react-redux';
 import renderer from 'react-test-renderer';
 import { Tag } from 'store/store-types';
-import { store } from 'store/store';
+import { ProviderForTesting } from 'store';
```
mt-xing: Is this supposed to be here?

SamChou19815: Yes, it's using absolute import.

PR #328, opened 2019-10-20
Start to unify usage of current time for deterministic testing

SamChou19815:
### Summary

Last time I tried to add more snapshot tests to increase the test coverage of samwise. The journey stops at `FutureView` because it uses `new Date()` all over the place. Using `new Date()` prevents effective snapshot testing because a snapshot taken on date A will fail on date B.

Unifying the usage of current time is a necessary step to solve the issue. If we can shield all usage of current time behind a framework, then we can easily dig some hole in the framework for overriding the system time with our hardcoded time for deterministic testing, as shown in this diff.

Unifying the usage also has another advantage: when we plan to support more specific time like `2019-10-30 17:00` in the future, we don't have to look all over the codebase to hunt down the usage of `new Date()` because it has already be resolved when we are pushing more deterministic time.

Surprisingly, it has the third advantage: now we can get rid of `react-live-clock` which does not have a typescript type definition. One less eslint warning!

This diff does not plan to solve all problems. More practices are left to our amazing members!

### Test Plan

- Added snapshot tests for the entire FutureView
- Manual Test: open the page, change your system time, now the live clock and future view layout should change

### DIFF
--- a/frontend/src/components/TaskView/FutureView/FutureViewDayTaskContainer.tsx
+++ b/frontend/src/components/TaskView/FutureView/FutureViewDayTaskContainer.tsx
@@ -4,7 +4,6 @@ import { CalendarPosition, FloatingPosition } from '../../Util/TaskEditors/edito
 import FutureViewTask from './FutureViewTask';
 import styles from './FutureViewDayTaskContainer.module.css';
 import { useWindowSizeCallback } from '../../../hooks/window-size-hook';
-import { error } from '../../../util/general-util';
 import { State } from '../../../store/store-types';
 import { createGetIdOrderListByDate } from '../../../store/selectors';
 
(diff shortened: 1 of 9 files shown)


frontend/src/components/TaskView/FutureView/__snapshots__/index.test.tsx.snap
```diff
@@ -0,0 +1,3759 @@
...
+        title="More"
+        width="1em"
+      >
+        v.svg
+      </svg>
+    </span>
+    <h3
+      className="Title"
+    >
+      1/10/27 - 1/23/27
```
SamChou19815: `currentTime` is indeed overridden!

frontend/src/components/TaskView/FutureView/FutureViewDayTaskContainer.tsx
```diff
@@ -42,7 +41,10 @@ function FutureViewDayTaskContainer(
 
   // Subscribes to it, but don't use the value. Force rerender when window size changes.
   useWindowSizeCallback(() => {
-    const containerNode = containerRef.current || error();
+    const containerNode = containerRef.current;
+    if (containerNode == null) {
+      return;
+    }
```
SamChou19815: Change to unblock snapshot test

frontend/src/components/TitleBar/index.tsx
```diff
@@ -1,6 +1,5 @@
 import React, { ReactElement } from 'react';
-// @ts-ignore this module does not have a ts definition file.
-import Clock from 'react-live-clock';
```
SamChou19815: Bye!

PR #331, opened 2019-10-21
Nuke undo

pihucode:
### Summary

Removed undo and added a confirmation prompt when user deletes a non-forked one-time task.

- [x] nuked undo
- [x] added confirmation prompt before deleting task
- [x] made changes to modal button style and moved the 'cancel' button to the left

### Test Plan 
remove onetime task modal


frontend/src/util/task-util.ts
```diff
@@ -237,16 +237,22 @@ function removeOneTimeTask(task: OneTimeTask): void {
...
+    });
+  } else {
+    promptConfirm(
+      'Do you really want to remove this forked task? The removed task cannot be recovered.',
+    ).then((confirmed) => {
+      if (confirmed) {
+        removeTask(task);
+      }
+    });
   }
```
mt-xing: Is it necessary to have both branches of this if statement? Can we just have a single conditional to change the string passed to the dialog box?

pihucode: Then would the message be "Do you really want to remove this task? The removed task cannot be recovered." for both forked and non-forked tasks?

mt-xing:
Why is cancel on the left for a repeating task but on the right for a normal one?

image

image

Also probably want a bit more horizontal padding on that last confirm button; it looks a bit squished as is.

mt-xing:
The staging build still shows the cancel button on the wrong side; can you check if it's happening on your end?

image

pihucode:
@mt-xing When I run `yarn start` my local server shows the updated changes, but the staging site doesn't reflect those changes.

Issue #327, opened 2019-10-18
Save button in future view editor can go off screen


mt-xing:
**Describe the bug**
A task with a lot of subtasks can push the save button off screen

**To Reproduce**
Make a task with a lot of subtasks. Open the task in future view. Edit something like the date. Try to click save.

**Screenshots**

Note that you can't even see the save button b/c it has been pushed off the bottom of the screen:

![image](https://user-images.githubusercontent.com/6147405/67096523-6aba0b80-f186-11e9-92d7-65adc4fa571b.png)

**Environment (optional)**
Windows 10, Chrome

ptwu:
Fixed in #337

PR #337, opened 2019-10-26
Fix off screen save button in FutureView editor

ptwu:
### Summary 

Added overflowY auto along with a maxHeight style to the editorPosDiv in future view which allows the user to scroll and see an unlimited number of subtasks, and be able to click the save button even with a large number of subtasks.

- [x] fixed issue #327 

### Test Plan 

efWeTLK


ptwu:
> Small question:
> Is it possible to do it in css?
> I think the reason why some of the css settings are done in js here is because we want to use some component position and location information that cannot be statically decided.

That makes sense. I admit that I got a bit too excited when I found the right component, just committed a change that adds it too the css of FloatingTaskEditor.

PR #342, opened 2019-10-27
Increase test coverage of the frontend codebase

SamChou19815:
### Summary

Add a lot of snapshot tests for components. Increased test coverage from 26% to 39%.

We have to set `TZ=UTC` for tests to ensure consistency across all environments for testing.

### Test Plan

The new tests are all passing!


SamChou19815:
The CI environment and I are not in the same timezone :(
Trying to find a way to fix that...

PR #344, opened 2019-10-28
Repeat task today

### LINKED ISSUE #341: Repeating Tasks Don't Start Today
**Describe the bug**
Repeating tasks are never added to today's tasks, even when today is ticked off on the "Repeats every week on" section. 

**To Reproduce**
Create any repeating task and make sure you click the day of the week for today. Add the task and bear witness to the lack of task adding to today!

Additionally, if you try to pin it to focus when creating it, it doesn't add it to focus!

**Screenshots**
image
image

**Environment (optional)**
 - Browser chrome
 - OS: macOS

jason-tung:
### Summary 
When creating a task to add to firebase/render, create the task start time as 00:00:00 and the end time 23:59:59. This solves the issues with respect to inclusivity and exclusivity on either end of the repeating task. 

- [x] fixed #341

### Test Plan 
image
image

### Notes 

Semantics can be improved/made more clear for "stops after x weeks".
Additionally fixes an issue that was not posted: creating repeating tasks does not exclude the first day after the repeat task duration when rendering the tasks.

### DIFF
--- a/frontend/src/components/TaskCreator/DatePicker.tsx
+++ b/frontend/src/components/TaskCreator/DatePicker.tsx
@@ -357,8 +357,9 @@ export default function DatePicker(props: Props): ReactElement {
     } else {
       // TODO once database support exists, replace this with number of occurrances
       endDate = new Date(new Date(
-        new Date().getTime() + 1000 * 60 * 60 * 24 * 7 * internalDate.repeatEnd.weeks,
-      ).setHours(0, 0, 0, 0));
+        new Date().getTime() + 1000 * 60 * 60 * 24 * 7 * internalDate.repeatEnd.weeks
+        - 1000 * 60 * 60 * 24,
+      ).setHours(23, 59, 59, 999));
     }
 
     const repData: RepeatMetaData = {
--- a/frontend/src/components/TaskCreator/DatePicker.tsx
+++ b/frontend/src/components/TaskCreator/DatePicker.tsx
@@ -1,8 +1,9 @@
 import React, { ReactElement, SyntheticEvent, ChangeEvent } from 'react';
 import Calendar from 'react-calendar';
+import { useTodayLastSecondTime, useTodayFirstSecondTime } from 'hooks/time-hook';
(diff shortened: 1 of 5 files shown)


frontend/src/components/TaskCreator/DatePicker.tsx
```diff
@@ -356,13 +356,14 @@ export default function DatePicker(props: Props): ReactElement {
...
-      );
+      endDate = new Date(new Date(
+        new Date().getTime() + 1000 * 60 * 60 * 24 * 7 * internalDate.repeatEnd.weeks
+        - 1000 * 60 * 60 * 24,
+      ).setHours(23, 59, 59, 999));
     }
 
     const repData: RepeatMetaData = {
-      startDate: new Date(),
+      startDate: new Date(new Date().setHours(0, 0, 0, 0)),
```
SamChou19815: We started to move away from using `new Date()` everywhere in https://github.com/cornell-dti/samwise/pull/328?
Could you move it to the react hook instead?

frontend/src/components/TaskCreator/DatePicker.tsx
```diff
@@ -356,13 +356,14 @@ export default function DatePicker(props: Props): ReactElement {
       endDate = internalDate.repeatEnd.date;
     } else {
       // TODO once database support exists, replace this with number of occurrances
-      endDate = new Date(
-        new Date().getTime() + 1000 * 60 * 60 * 24 * 7 * internalDate.repeatEnd.weeks,
-      );
+      endDate = new Date(new Date(
+        new Date().getTime() + 1000 * 60 * 60 * 24 * 7 * internalDate.repeatEnd.weeks
+        - 1000 * 60 * 60 * 24,
+      ).setHours(23, 59, 59, 999));
```
SamChou19815: Better move this to [datetime-utils](https://github.com/cornell-dti/samwise/blob/master/frontend/src/util/datetime-util.ts) and accompany it with tests.

Issue #349, opened 2019-11-03
Lots of tasks on same date: calendar gets cut off


meganyin13:
**Describe the bug**
When you have a lot of tasks on one date, if you try to click calendar on last task, it gets cut off.

**To Reproduce**
Create many tasks on same date such that it overfills vertical width. On the last one, open calendar, it is cut off.

**Screenshots**
Calendar gets cut off here:
![image](https://user-images.githubusercontent.com/20008134/68093298-24d99480-fe62-11e9-9584-52cd1cb9f9a1.png)

**Environment (optional)**
 - Browser Chrome
 - OS: macOS

**Additional context (optional)**

SamChou19815:
Fixed in #356

PR #356, opened 2019-11-10
Fix UI bugs: Calendar cutoff with large # of tasks, save button disappear

### LINKED ISSUE #348: Adding multiple subtasks: unable to see save button
**Describe the bug**

When you add multiple subtasks to a task in future view, cannot scroll down and see save button.

**To Reproduce**
Create a task (repeated or one-time). Try adding a lot of subtasks such that it goes over the bottom of the screen. Try scrolling, you won't be able to find the save button.

**Screenshots**

Cannot scroll down to see save button here:
image

**Environment (optional)**
 - Browser Chrome
 - OS: macOS

ptwu:
### Summary 

This pull request originally changed the max-height attribute of the FloatingTaskEditor to make sure that subtasks that even go past the viewport still allow you to save and access the calendar.

@SamChou19815 made a more involved fix to detect change in height with the save button in the reducer which is committed below

- [x] fixes #348 and #349

### Test Plan 

OR1bQrv - Imgur

image
^ More Tasks FloatingView doesn't get cut off anymore by the left panel


frontend/src/components/Util/TaskEditors/FloatingTaskEditor.module.css
```diff
@@ -4,7 +4,7 @@
   z-index: 8;
   box-shadow: 0 5px 12px 0 rgba(0, 0, 0, 0.3);
   transition: width 0.1s linear;
-  max-height: 400px;
+  max-height: 200px;
```
SamChou19815: I'm surprised that decreasing the height can solve the problem, instead of increasing it.

ptwu: Yeah, the decreased height makes the editor overflow if there are enough subtasks for it to go below the viewport, and then the user can use the scrollbar to change the due date / save.

ptwu:
> But now tasks that should fit on screen no longer fit.
> 
> image

here's a gif of my new fix:
QvTzC8i - Imgur

SamChou19815:
@mt-xing made the actual fix to the hidden save button issue. See my commit message above.

Issue #340, opened 2019-10-27
New Task Editor Design


mt-xing:
Update `TaskEditor` to look like the designs on Zeplin

![image](https://user-images.githubusercontent.com/6147405/67640412-d6138400-f8d0-11e9-8f15-bbc600dc4242.png)

If a task repeats on one or two days, list the days it repeats. If there's 3 or more days a task repeats, just say "repeats x days a week" where x is the number of days.

Blocks #313

SamChou19815:
Fixed in #357

PR #357, opened 2019-11-10
Add repeat frequency header

pihucode:
### Summary 

Tasks which repeat for two days or less per week have the days for which they repeat listed on a frequency header. Tasks which repeat for three or more days per week have the number of days that they repeat per week displayed on the frequency header.

- [x]  New task editor design #340 

### Test Plan 
repeatfreqheader single day
repeatfreqheader 2 days
repeatfreqheader 7 days

### Notes 

Possible contrast issues with light colors.


SamChou19815:
Used react-redux to ensure it also gets displayed in forked instances.

mt-xing:
There definitely shouldn't be a scrollbar here:

image

SamChou19815:
Will fix. Looks like my css hack needs more work

PR #363, opened 2019-11-24
Drag and Drop (N Days)

meganyin13:
### Summary

Added drag and drop functionality for each day in FutureView. 

This pull request is the first step towards implementing feature Drag and Drop

- [x] implemented Drag and Drop (same day)

### Test Plan

Create 3-5 one time tasks on the same day. Click and hold on the three dot draggable icon next to each task and drag it to a different position in between other tasks. Ensure that when this happens, the tasks move to make space for the task you're dragging. Release, and ensure that the tasks are reordered. 

### Notes

Known issue: doesn't work when dragging a task to the bottom of list of tasks in a day.


ptwu:
testyboy

frontend/src/components/TaskView/FutureView/FutureViewDayTaskContainer.tsx
```diff
@@ -54,23 +59,57 @@ function FutureViewDayTaskContainer(
...
+      )}
+    </Draggable>
   ));
   if (isInMainList) {
-    const style = { overflow: 'hidden' };
+    const style = {};
     return (
-      <div className={styles.Container} style={style} ref={containerRef}>{taskListComponent}</div>
+      <DragDropContext onDragEnd={onDragEnd}>
+        <Droppable droppableId="onetestyboy">
```
SamChou19815: It's OK to have fun in the code, but you should pick some name that is both
- fun
- at least reveal some information about the object you are modeling.
Here is a good example in my opinion: https://github.com/cornell-dti/samwise/blob/37eca33929116f626b806c08d1da266384c23811/frontend/src/util/tag-util.ts#L6
`onetestyboy` here simply adds confusion.

ptwu: image
does this fit the criteria? 🤔

ptwu: Droppable ID has been changed to the boring "future-view-task-droppable"

PR #365, opened 2019-11-24
Add Canvas calendar field

pihucode:
### Summary

Add a field at the bottom of Settings that allows users to add their Canvas calendar feed link. This calendar feed is able to be unlinked. A help link is included below the field in case users have trouble finding their Canvas calendar feed link.

- [x] Add Canvas calendar field to settings page

### Test Plan 

canvas calendar field link
canvas calendar field unlink

### Breaking Changes

Adds a _canvasCalendar_ field to documents in _samwise-settings_ in Firestore.

### DIFF
--- a/frontend/src/components/TitleBar/Canvas/CanvasCalendar.tsx
+++ b/frontend/src/components/TitleBar/Canvas/CanvasCalendar.tsx
@@ -51,15 +51,15 @@ function CanvasCalendar({ settings }: Props): ReactElement {
 
         <div
           className={settingStyles.SettingsButton}
-          style={{
-            marginTop: '12px',
-            display: linked ? 'block' : 'none',
-          }}
+          style={{ display: linked ? 'block' : 'none' }}
         >
-          Your Canvas calendar feed is linked.
+          <p style={{ wordBreak: 'break-all' }}>
+            Your Canvas calendar feed is linked.
+            <br />
+            {canvasCalendar}
+          </p>
           <button
             type="button"
-            style={{ marginLeft: '12px' }}
             onClick={removeCanvasiCal}
             title="Remove Canvas iCal Link"
           >
--- a/frontend/src/components/TitleBar/Canvas/CanvasCalendar.tsx
+++ b/frontend/src/components/TitleBar/Canvas/CanvasCalendar.tsx
@@ -1,7 +1,8 @@
 import React, { ReactElement, Component } from 'react';
-import styles from '../Settings/SettingsPage.module.css';
+import settingStyles from '../Settings/SettingsPage.module.css';
 import { settingsCollection } from '../../../firebase/db';
 import { getAppUser } from '../../../firebase/auth-util';
+import styles from './CanvasCalendar.module.css';
 
 type State = {
   canvasCalendar: string;
@@ -15,12 +16,11 @@ export default class CanvasCalendar extends Component<{}, State> {
   };
 
   componentWillMount(): void {
-    const self = this;
     settingsCollection().doc(getAppUser().email)
       .get().then((doc) => {
         const userSettings = doc.data();
         if (userSettings) {
-          self.setState({ linked: !(userSettings.canvasCalendar === null) });
+          this.setState({ linked: !(userSettings.canvasCalendar === null) });
         }
       });
   }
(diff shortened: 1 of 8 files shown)


frontend/src/components/TitleBar/Canvas/CanvasCalendar.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+type State = {
+  canvasCalendar: string | null;
+  linked: boolean;
+}
+
+export default class CanvasCalendar extends Component<{}, State> {
+  public state: State = {
+    canvasCalendar: '',
+    linked: readCanvasCalendar() != null,
```
SamChou19815: This is problematic here. You can potentially lose sync with remote canvas calendar field.
You should subscribe to redux store instead.

Subscription code:
https://github.com/cornell-dti/samwise/blob/ee0c0e22ad80895322528003f4f732e9a3270b7b/frontend/src/components/TitleBar/Settings/SettingsPage.tsx#L130-L131

Usage:
(Subscribed props becomes normal props.)
https://github.com/cornell-dti/samwise/blob/ee0c0e22ad80895322528003f4f732e9a3270b7b/frontend/src/components/TitleBar/Settings/SettingsPage.tsx#L62-L69

frontend/src/components/TitleBar/Canvas/CanvasCalendar.tsx
```diff
@@ -0,0 +1,77 @@
...
+import settingStyles from '../Settings/SettingsPage.module.css';
+import styles from './CanvasCalendar.module.css';
+import { setCanvasCalendar } from '../../../firebase/actions';
+
+type Props = { readonly linked: boolean }
+
+type State = { canvasCalendar: string | null }
+
+export default class CanvasCalendar extends Component<Props, State> {
+  public state: State = { canvasCalendar: '' };
```
SamChou19815: Why don't we simply pass `canvasCalendar` down and don't keep any state in this component?

pihucode: Yep, not having a state would be best. The iCal link that the user inputs is stored in `canvasCalendar` which is passed to `setCanvasCalendar()`. I'm not sure how to get the current value in a form field without storing it in a state.

frontend/src/components/TitleBar/Canvas/CanvasCalendar.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  handleChange = (e: React.ChangeEvent<HTMLInputElement>): void => {
+    this.setState({ canvasCalendar: e.target.value });
```
SamChou19815: Use `e.currentTarget.value`

frontend/src/components/TitleBar/Canvas/CanvasCalendar.tsx
SamChou19815: We seems to be hit by the same `TaskEditor` problem here. Duplicating the link might be necessary.
However, currently, the user can't see the link unless they unlink first.

pihucode: Then should I display the iCal link below the "Your Canvas calendar feed is linked" text?

SamChou19815: Looks good to me

frontend/src/store/store-types.ts
```diff
@@ -84,6 +84,7 @@ export type PartialMainTask = Partial<MainTask>;
  * The type of user settings.
  */
 export type Settings = {
+  readonly canvasCalendar: string | null;
```
SamChou19815: The type should be `string | null | undefined` because some db objects might not have this field. Since you are using `!= null` check, you automatically get the `undefined` case handled.
However, let's be precise here to avoid future confusion. You could either do `readonly canvasCalendar: string | null | undefined;` or eliminate the `undefined` case in firebase listener.

pihucode: Okay, changed the type to be `string | null | undefined`

mt-xing:
Should there be some clickable save button next to the field? Pressing Enter may not be too intuitive for users, especially when every other field autosaves.

pihucode:
> Should there be some clickable save button next to the field? Pressing Enter may not be too intuitive for users, especially when (almost) every other field autosaves.

@mt-xing I can add that tomorrow. Should the button be beside (to the right) the text input or below it?

mt-xing:
> @mt-xing I can add that tomorrow. Should the button be beside (to the right) the text input or below it?

To the right makes more sense to me, but whatever works for you.

PR #374, opened 2019-12-04
New Landing Page

mt-xing:
### Summary 

Created the new landing page as per the designs on Zeplin. Currently there is only one login with Google button, as we don't have time to work around the limitations of the login component before this semester's deadline.

### Test Plan 

Look at it.

screencapture-staging-pr-samwise-today-374-2019-12-03-22_17_06

Also works decently well on narrow screens:

screencapture-localhost-3000-2019-12-03-21_59_56

### Notes 

In addition to needing to be updated with a second Login button, the loading screen will briefly flash for a split second before the landing page loads. We can fix this eventually, but it works for now.


SamChou19815:
> the loading screen will briefly flash for a split second before the landing page loads

Both staging and prod have the same problem, so we are not making anything worse.

mt-xing:
> `frontend/src/assets/splash/screenshot.png` is 1.58MB. We might need to compress it. @mt-xing

Fixed in #375

PR #375, opened 2019-12-04
Compress screenshot.png

mt-xing:
### Summary 

Drastically reduced the filesize of `screenshot.png` by reducing the size of the image. Opening the original picture revealed that it was a small picture that had been upscaled and did not actually have 4000 horizontal pixels. As such, shrinking the image maintains quality while reducing file size over 80%.

### Test Plan 

The landing page still looks fine to me.


ptwu:
you could also use smart lossy compression to further decrease the file size to below 50 KB. (i used https://compressor.io/)

mt-xing:
> you could also use smart lossy compression to further decrease the file size to below 50 KB. (i used https://compressor.io/)

I'm already lossy compressing by throwing away 75% of the pixels; I'm not sure if we want to compress it any more.

PR #385, opened 2019-12-07
RFC: Extract some common firebase code

SamChou19815:
### Summary

This is a proof of concept of extracting common firebase code for both frontend and functions. This is more complex than I will hope. I will list sources of complexity below and explain my strategy to resolve them. I think some of my approaches will be controversial. I'm open to better plans, but my current plan is the best I can think of for now.

#### Problem: incompatible firestore between web frontend and admin.

The firestore from `firebase/firestore` (I'll call it F1 from now on) and from 'firebase-admin/firestore' (F2) are slightly different. Type of F1 and F2 are not assignable to each other. The slight difference in the types are very deep, so it's impossible to just to add `@ts-ignore` to a few places.

Here comes from my controversial approach. Copy paste some of firestore's type into `firebase/database`, remove the non-common parts and try to reconcile that between frontend and functions folder. Since TS uses structral subtyping, the passed in F1 and F2 can be both assigned to the common Firestore type and everything will still work.

Here is a danger of this approach: once firebase changes some of the definition, upgrading it would be expensive. However, we know these definitions tend to be stable, so I guess we won't run into too much problem.

#### Proof of concept migration

I migrated all firebase actions involves tags to the new common code architecture and migrated the rest of the code to use the new `database` object.
Right now, only the database file and order manager file are fully migrated.
I also made the functions folder use the new architecture to prove that it works.

### Test Plan

Everything still runs. Things type check!


common/src/firebase/common-actions.ts
```diff
@@ -0,0 +1,61 @@
...
+import OrderManager from './order-manager';
+import { ignore } from '../util/general-util';
+import { NONE_TAG_ID } from '../util/tag-util';
+
+type WithoutIdOrder<Props> = Pick<Props, Exclude<keyof Props, 'id' | 'order'>>;
+
+export default class Actions {
+  readonly orderManager: OrderManager;
+
+  constructor(private readonly getUserEmail: () => string, private readonly database: Database) {
```
SamChou19815: We need to delay `getUserEmail` evaluation since it is only available after auth.

common/src/firebase/common-actions.ts
```diff
@@ -0,0 +1,61 @@
...
+        s.docs
+          .filter((doc) => doc.data().type === 'TASK')
+          .forEach((doc) => {
+            batch.update(this.database.tasksCollection().doc(doc.id), { tag: NONE_TAG_ID });
+          });
+        batch.delete(this.database.tagsCollection().doc(id));
+        batch.commit().then(ignore);
+      });
+  };
+}
```
SamChou19815: This is not intended to be a complete replacement of all the actions code.
For example, the actions code in frontend has some analytics code, which is left over there.

PR #392, opened 2020-01-02
Dark Mode (internal-only)

mt-xing:
### Summary 

If the designers aren't going to give us a dark mode any time soon, then I'm making one myself.

This particular pull request does **not** give users the ability to actually switch to dark mode. It only displays the dark mode if the user's theme settings in Firestore reads `dark`, which right now can only be accomplished through manually editing the entry in the Firebase Console. We probably don't want to expose dark mode to the public until we get either a designer or PM's approval.

### Test Plan 

Look at it. It's beautiful. (still a WIP)

image


frontend/src/components/TaskCreator/index.tsx
```diff
@@ -48,11 +51,22 @@ const initialState = (): State => ({
...
 
+  private darkModeStyles: undefined | CSSProperties;
+
+  constructor(props: {theme: Theme}) {
+    super(props);
+    const { theme } = props;
+    this.darkModeStyles = theme === 'dark' ? {
+      background: 'black',
+      color: 'white',
+    } : undefined;
```
SamChou19815: `darkModeStyles ` can be out-of-sync with the props if you wrote it in this way.

frontend/src/components/TaskView/FutureView/FutureViewDay.tsx
```diff
@@ -162,3 +183,10 @@ export default function FutureViewDay(props: Props): ReactElement {
     </div>
   );
 }
+
+const Connected = connect(
+  ({ settings: { theme } }: State, ownProps: Props): Props & { readonly theme: Theme } => (
+    { ...ownProps, theme }
```
SamChou19815: Correct me if I'm wrong, but the logic here seems unnecessary. If you do
```tsx
const Connected = connect(
  ({ settings: { theme } }: State):  { readonly theme: Theme } => ({  theme }),
)(FutureViewDay);
```
it's equivalent to what you did here. `react-redux` automatically handles this common case.

PR #395, opened 2020-01-26
New release for SP20

SamChou19815:
### Summary 

This release contains

- partially finished drag and drop
- courses and exams for SP20

### Test Plan 

- play with drag and drop within the same date
- play with new courses


mt-xing:
> testinprod

I feel obligated to point out that this is a bad test plan

PR #404, opened 2020-02-01
Add an integration test for course info scraper

SamChou19815:
### Summary

I just thought about it today and found that we can easily write an integration test to test whether the latest course-info json is consistent with the one on the master. We need this check to ensure that our information is up-to-date.

This check proves to be necesarry, since I do found some inconsistencies, probably because some courses and exams are added. The course json has been updated to reflect the latest state.

You might think of this alternative: always building the new json at deploy time. This approach might work in theory, but is really dangerous in practice. Imagine Cornell suddenly changed the format of the exams page in a way such that the script will not crash but simply produce bad result, then the bad json will go into prod!

### Test Plan

See the CI job yml. Note that I used a tool called `jq` to normalize the json. It can remove potential inconsistency caused by json entries order and ensure that in case there is an inconsistency, the diff outputed is more readable.


SamChou19815:
:( There are still some timezone issues

Issue #408, opened 2020-02-05
Drag-and-Drop Within a Day in Future View Breaks Ordering in Focus View


mt-xing:
**Describe the bug**
Trying to reorder two tasks in future view within a day will break the ordering of stuff in focus view.

**To Reproduce**
Create three tasks, two due on one day and a third due on a different day, which I'll call a, b, and c, respectively.
Focus all three tasks.
In focus view, drag c so that it is placed between a and b.
In future view, assuming a comes before b, use drag-and-drop to place b before a.
Note that focus view is now in a different order.
What's worse, attempts to reorder focus view causes all types of wack, unexpected behavior.

Demo:
https://youtu.be/O24cDUP1aUw

**Environment (optional)**
Tested on production and staging. Can be reproduced in both.

PR #410, opened 2020-02-08
Drag and Drop N Days

meganyin13:
### Summary 

Allow drag and drop across n days.

WORK PLAN/ISSUES:
- [x]  allow drag and drop n days in 5D view
- [x]  allow drag and drop to day with no tasks
- [x]  bring back task reordering for drag and drop same day
- [x]  disable dragging ability for repeating task
- [x]  disable dragging ability for repeating task
- [x]  fix 2W and monthly view
- [x]  make dark mode work with drag and drop

This pull request implements the long-awaited feature Drag and Drop (single day + across N days)

- [x] implements Drag and Drop!

### Test Plan 

Make at least 3 tasks on at least 2 different days. For clarity, number each task and label the day the task starts on. Drag these tasks to different dates, ensure they appear where you expect. Also try dragging these tasks to dates with no tasks. Create a repeated task and ensure that you are unable to drag it.

Ensure that single day drag and drop is unaffected by dragging tasks within one day.

### Known Issues
WILL BE FIXED IN NEXT ITERATION:
- Dragging and dropping about a repeated task on a single day causes some wacky UI behavior. 
- Need to use caching to fix the flash that occurs after dropping. 
- We are CORRUPTING the database. FutureView order should be independent of FocusView. #408.


frontend/src/components/TaskView/FutureView/FutureViewDayTaskContainer.module.css
```diff
@@ -1,3 +1,4 @@
 .Container {
   height: calc(100% - 66px);
+  min-height: 216px;
```
SamChou19815: This is problematic. Things might only work for certain screen size.

ptwu: we can try to calculate the container height instead, or an easier way to deal with this is to just disable drag and drop for 2W and monthly views

mt-xing:
You're not supposed to be able to change the due date of Canvas tasks, yet you can still drag them into different days.

mt-xing:
Dark mode is broken

mt-xing:
Why do Canvas tasks have a repeating icon?
image

mt-xing:
Whenever I drag a task to a new day, it always shows up at the top of that day, no matter where I drop it.

Also dark mode is still broken.

PR #422, opened 2020-02-16
Canvas Tasks Off By One

jason-tung:
### Summary 
Fixes issue where tasks due at 11:59 PM were being shown on the wrong date by writing a parser that handles cases where the ICal given by Canvas is not sufficient to accurately assign a date to a task.

Goes line by line to scan for events and match the following lines with corresponding fields like dates and names for each calendar. Next matches these fields to actual firebase object and uploads it to database.

### Test Plan 
Ran unit testing against ical-parser as well as the date helper functions. Running the ical-functions script displays updates firebase information to expected values and tasks appear on the correct dates.

### Notes 
The way the parsing is done is somewhat sloppy and some of my helper functions may be better suited to be combined into a util/date-time module later on.


SamChou19815:
@jason-tung could you fix the linter and build first?

functions/src/iCalFunctions.ts
```diff
@@ -1,83 +1,87 @@
 /* eslint-disable no-await-in-loop */
-import { fromURL } from 'ical';
-import OrderManager from 'common/lib/firebase/order-manager';
-import database from './db';
+import fetch from 'node-fetch';
+import icalParse from './ical-parser';
+import { settingsCollection, tasksCollection }
+  from '../../frontend/src/firebase/db';
+import getOrder from '../../frontend/src/firebase/order-manager';
```
SamChou19815: Use
```suggestion
import OrderManager from 'common/lib/firebase/order-manager';
```

PR #423, opened 2020-02-17
RFC: Major Refactor of Task and SubTask in Redux Store

SamChou19815:
### Summary 

This diff is the first step for my planned refactoring of frontend representation of tasks.
This diff tries to merge main task and subtask together in redux store, rather than separating them into two different maps.

#### Historical Context

Let me first explain why main task and subtasks are separate in the first place: it has something to do with firebase realtime update. We listen to two different collections: `samwise-task` and `samwise-subtask`. It's guaranteed that these two different collection will return the latest update at different times.
During initial fetch, we can hack around it by first fetching tasks, then subtasks. However, we still have no control over the subsequent update orders. Therefore, it seems very convenient to just model the redux store based on the database, and let the rest of the application code handle the logic of merging the information from these two maps together.

#### Why It Is Bad

The above approach seems very convenient to implement while I was migrating the codebase to firebase almost a year ago, but it puts a huge burden on the rest of the application code. Some code that does not require `react-redux`'s `connect` suddenly needs them. We also end up with a lot of code like

```js
children.forEach(id => {
  const subTask = subTasks.get(id);
  if (s == null) {
    return;
  }
  // do something with subTask
})
```

, which won't exist if subtask is together with main task.

#### New Approach

What if we try to merge them together, regardless of the complexity of the reducer? I tried by starting to changing `Task` type to have an array of `SubTask` instead of a set of id. It turns that this move simplified a lot of code than I originally expected. Therefore, I think the approach is profitable and go on with the rest of the refactoring.

So now we are going to put tasks and subtasks together, what are the problems we can have? It turns out that there are only two:

- When tasks arrives before subtasks, some tasks expect a subtask with id `foo`, but the update of subtask with id `foo` has not come yet.
- When subtasks arrives before tasks, we don't know where to stick into some subtasks.

In the code, I refer to these problems as 

- missing subtask
- orphan subtask

Once the problem is clearly formulated, the solution is obvious: `missingSubTasks` and `orphanSubTasks` are now tracked by the redux store.

And we can formulate an invariant below:

> If all updates have been performed and database is not corrupted, all tasks object should contain all of its subtasks, and `missingSubTasks` and `orphanSubTasks` are empty.

The job of the reducer is then to enforce the invariant. It's a little hard to reason about the current state of the app, so I wrote down some comments in the code to discuss all the cases.

#### How is it Related to Future Plan

In the future, I plan to make frontend `Task` to have a structure like

```typescript
type CommonTask<M> = {
  readonly id: string;
  readonly order: number;
  readonly name: string;
  readonly tag: string;
  readonly complete: boolean;
  readonly inFocus: boolean;
  readonly subTasks: SubTask[];
  readonly metadata: M
}
type Task =
  | CommonTask<{ readonly date: Date }> // the current one-time task
  | CommonTask<RepeatingTaskMetaData> // the current repeating task
  | CommonTask<GroupTaskMetadata> // the potential group task
  | CommonTask<CanvasMetadata> // a more principled version of canvas task
  | CommonTask<ExamData> // auto-generated exam tasks, don't explicitly exist in DB
```

I believe this approach can make our codebase scale when more different types of tasks are introduced, because we only need to pattern matching on one field instead of the entire task. Some presentation only components can then just passing those metadata around without looking at them.

In order for this vision to come true, main task and subtask would better be stored together.

### Test Plan 

- Played tested the app. Nothing seems to break now.
- Added unit test of reducer to test the tricky patch behavior.

### Notes

The most interesting changes are in `store-types.ts` and `reducers.ts`. All other changes are simple brainless refactoring.


mt-xing:
I'm going to hold off on approving this since the notes say not to approve yet. Ping me when ready.

SamChou19815:
@mt-xing I added reviewers just to collect early feedback.

SamChou19815:
Now this PR is fully ready.
@meganyin13 @mt-xing @ptwu @pihucode @jason-tung

PR #428, opened 2020-02-23
Make build and test in frontend work across different OS

SamChou19815:
### Summary

Instead of hardcoding envvar in the script, put them in a env file and load it by cross-platform node.js code.

- Now `yarn build:stage` should work cross-platform
- `yarn test` used to set `TZ=UTC`, but it only works on unix based OS. Only one trivial test is affected by this, so I just removed the test.

### Test Plan

Testing it locally. It works. See CI log to verify that it also works in CI.
Can a reviewer on Windows help me test whether it works on Windows?

### Notes

Inspired by https://create-react-app.dev/docs/deployment#customizing-environment-variables-for-arbitrary-build-environments.


SamChou19815:
Removed the affected test

PR #433, opened 2020-03-01
[WIP] Assign tags to Canvas tasks

pihucode:
### Summary 
When an iCal feed is linked to a user's calendar, the imported Canvas tasks have the default "None" tag. We want to choose a suitable tag from the user's current list of tags to assign to their Canvas tasks.

Names of class tags from imported classes have the form "CS 4820: Intro Analysis of Algorithms". Canvas task names have the form "HW 4 [ENGRD2700]". Regex is used to extract course codes (for example, "CS4820") from the Canvas task and tag names. 

Compare the course codes (for example "CS4820" and "ENGRD2700"). Course codes are case-sensitive. If the course codes match, then the tag id is assigned to the Canvas task. If the user has their own custom tag named "CS 2110" and also has a class tag "CS 2110", the custom tag takes precedence.

- [x] Tags are assigned to Canvas tasks when the task is pinned to focus view or if the task editor is opened.

### Test Plan
- Open the task editor for a Canvas task, or pin the Canvas task to the focus view. A tag matching the course code should automatically be assigned to the task.
- If the tag for a Canvas task is changed, reopening the task editor or pinning it to focus view will reset the tag to the assigned tag.

### Notes
- Canvas tasks should have suitable tags automatically assigned to them, but assignCanvasTag() is triggered only when the EditorHeader for task editors is rendered or when tasks in focus view are rendered.

### DIFF
--- a/frontend/src/components/Util/TaskEditors/TaskEditor/EditorHeader.tsx
+++ b/frontend/src/components/Util/TaskEditors/TaskEditor/EditorHeader.tsx
@@ -1,6 +1,7 @@
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useEffect } from 'react';
 import Calendar from 'react-calendar';
 import { Tag, RepeatMetaData } from 'common/lib/types/store-types';
+import { getCanvasTag } from 'firebase/actions';
 import styles from './index.module.css';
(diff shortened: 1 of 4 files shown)


mt-xing:
Going to close this for now as we move onto group tasks.

PR #436, opened 2020-03-03
Uncorrupt new tasks in database

### LINKED ISSUE #408: Drag-and-Drop Within a Day in Future View Breaks Ordering in Focus View

meganyin13:
### Summary 

Currently, drag and drop in FutureView is using the order property that was intended just for FocusView. However, if you focus 2 tasks from one day and then another task from another day, and drag the task from the second day in between the other two in FocusView, further drag and drop reorderings in FocusView become messed up (#408). The order property is no longer unique for all tasks (the invariant required for FocusView drag and drop).

~~To fix this, we will add a new field `future_view_order` to every task that will be unique for each day. If there are _k_ canvas and repeated tasks, those will be the first _k_ orders for that day. We will also write unit tests to ensure that for a single day, there are no two tasks on that day with the same `future_view_order`. This will decouple the FutureView drag and drop from the FocusView.~~
We chose to no longer add a new field due to directions from the PM and focus on implementing new features in the timeline instead. The current status of this pull request focuses on fixing the corrupted new tasks as detailed in #408.

Work Plan:

- [x] fixed #408 
- [x] unit tests for order

### Test Plan 

Attempt to reproduce #408. See that the bug in drag and drop is no longer.


SamChou19815:
> However, if you focus 2 tasks from one day and then another task from another day, and drag the task from the second day in between the other two in FocusView, further drag and drop reorderings in FocusView become messed up (#408).

@meganyin13 Is this a bug of focus view or future view drag and drop? If it's the bug of focus view, then we need to fix the focus view. If it's the bug of future view, then I'm confused. Drag and drop between days will only change the date field. It will never corrupt the order field.

Anyway, I recommend figuring out the root cause of the problem before adding a new field.

meganyin13:
The bug is in FutureView. In #408, Michael found the bug by manipulating drag and drop in FocusView, but you can actually replicate the same issue by simply making task A on day 1, task B on day 2, and task C on day 1 in that order. `order` is the order in which the tasks are created, so task A has `order: 0` (or the lowest order), B has `order: 1`, C has `order: 2`. Now in FutureView, when you swap A and C (which are on the same day) A will have `order: 1` and C will have `order: 0`. But B also has `order: 1` and this breaks FocusView since FocusView assumes all orders are unique. 

The root cause is that **within a single day there may be breaks in the order**. But we reorder tasks assuming there aren't.

In `computeReorderMap()`, for all elements that aren't the source element, we increment or decrement their order by 1 depending on if it wants to go to earlier or later places. **This assumes there are no breaks in the order**. In the example above, A and C have orders 0 and 2. There is a break. Dragging C above A makes A have `order: 0+1` and C have `order: 0`. So before assigning new orders, we need to keep track of all orders that are available and reassign based on that.

SamChou19815:
@meganyin13 does the current future view implementation calls `computeReorderMap`? If it does, then why does the problem exist? Because FocusView is also dealing with a possible non-continuous list of orders.

meganyin13:
@SamChou19815 The issue doesn't exist for FocusView because FocusView calls `computeReorderMap()` on all tasks regardless of whether they are focused or not. So all tasks get reordered. For FutureView, we only call `computeReorderMap()` on tasks within that day.

SamChou19815:
@meganyin13 so if you call `computeReorderMap` for all tasks in `FutureView`, doesn't it solve all problems?

PR #437, opened 2020-03-03
Task Type Refactor

SamChou19815:
### Summary

This diff continues the effort made in #423. In this diff, we are finally able to group the content of the task together, and factor all one-time task and repeating task specific stuff to `TaskMetadata`. In the future, we can add more stuff to task metadata without greatly impacting the rest of our codebase.

This diff is intended to be the first step of cleaning up our frontend representation of task. It does not plan to cleanup the rest of the code. The only goal of this diff is to refactor the type, make the app compile, and ensure nothing breaks.

The only potentially controversial change is to move `date` into `TaskMetadata`. While it might seem odd because a task has a date is so central to the app, I believe it is necessary because the behavior is fundamentally different for one-time and repeating task. Giving repeating task a dummy date will make buggy code that makes bad assumptions of date escape through the type checker.

### Test Plan

- [x] Changed existing tests due to type refactoring.
- [x] Manually played with the app. Everything else still seems to work.

### Note

To ease migration, I did not touch the types of backend task objects at all. The compatibility is resolved in the `listener.ts` file.


common/src/types/store-types.ts
```diff
@@ -23,78 +23,59 @@ export type SubTaskWithoutIdOrder = Pick<SubTask, 'name' | 'complete' | 'inFocus
...
-  readonly date: D;
-  readonly complete: boolean;
-  readonly inFocus: boolean; // Whether the task is in focus
-  readonly children: readonly SubTask[];
-};
-
-type FlexibleCommonTask = CommonTask<Date | RepeatMetaData>;
-
-export type OneTimeTask = CommonTask<Date> & {
+// TODO: refactor ical task to a separate category
```
ptwu: Are iCal tasks not also one time tasks?

SamChou19815: The hope is that iCal tasks are in a completely separate category, so that pattern matching mechanism will force us not to forget special-casing different behavior. e.g. disable edit.

For example, we might want something like this:
```typescript
function canBeEdited(metadata: TaskMetedata) {
    switch (metadata.type) {
        case 'ONE_TIME':
            return true;
        case 'I_CAL':
            // we will never forget to consider correctness for iCal tasks
            return false;
        // other cases ...
    }
}
```

ptwu: makes sense, the rest of the PR looks good to me

PR #466, opened 2020-04-18
updated firestore types for group tasks

jason-tung:
### Summary 
Added optional groupfield to onetimetask and onetimetask metadata 
Added group type to firestore-types

### Test Plan 

- [x] Will let PR deploy and play with app on staging to ensure that app is working and none of the associated functionality like uneditable canvas tasks breaks.

- Database schema change (anything that changes Firestore collection structure)
Added a samwise-groups that tracks its name : string , members : string[], and deadline: Date

### DIFF
--- a/common/src/types/firestore-types.ts
+++ b/common/src/types/firestore-types.ts
@@ -47,10 +47,15 @@ export type FirestoreOneTimeTask = FirestoreCommonTask & {
   readonly type: 'ONE_TIME';
   readonly date: Date | firestore.Timestamp;
   readonly icalUID?: string;
-  readonly group?: string;
 };
 
-export type Group = {
+export type FirestoreGroupTask = FirestoreCommonTask & {
+  readonly type: 'GROUP';
+  readonly date: Date | firestore.Timestamp;
+  readonly group?: string;
+}
+
+export type FirestoreGroup = {
   readonly name: string;
   readonly members: string[];
   readonly deadline: Date;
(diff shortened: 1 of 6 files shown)


common/src/types/firestore-types.ts
```diff
@@ -47,8 +47,15 @@ export type FirestoreOneTimeTask = FirestoreCommonTask & {
   readonly type: 'ONE_TIME';
   readonly date: Date | firestore.Timestamp;
   readonly icalUID?: string;
+  readonly group?: string;
 };
```
SamChou19815, line 51: Do not nest this inside one time task. Create another type called `GroupTask`, and add that as a variant in `FirestoreTask`

common/src/types/firestore-types.ts
```diff
@@ -47,8 +47,15 @@ export type FirestoreOneTimeTask = FirestoreCommonTask & {
...
 
+export type Group = {
+  readonly name: string;
+  readonly members: string[];
+  readonly deadline: Date;
+}
+
```
SamChou19815, line 58: To be consistent, all firestore types should be prefixed with `Firestore`

common/src/types/store-types.ts
```diff
@@ -28,6 +28,7 @@ export type OneTimeTaskMetadata = {
   readonly type: 'ONE_TIME';
   readonly date: Date;
   icalUID?: string;
+  group?: string; // documentid for associated group
```
SamChou19815, line 31: Same here. Create a new type for group tasks

common/src/types/firestore-types.ts
```diff
@@ -49,6 +49,19 @@ export type FirestoreOneTimeTask = FirestoreCommonTask & {
   readonly icalUID?: string;
 };
 
+export type FirestoreGroupTask = FirestoreCommonTask & {
+  readonly type: 'GROUP';
+  readonly date: Date | firestore.Timestamp;
+  readonly group?: string;
```
SamChou19815, line 55: group field shouldn't be optional. Same below

PR #476, opened 2020-04-18
Cross-Platform Deployment Script

mt-xing:
### Summary 

Rewrite the function deployment script in node.js, so that it runs on all OSes consistently.

### Test Plan 

Ran it on Windows and it deployed to staging okay.


mt-xing:
Actually let me change one last thing

Issue #483, opened 2020-04-29
Broken staging page layout


SamChou19815:
**Describe the bug**

- Cannot scroll in focus view
- Extra right margin

**To Reproduce**

Visit staging site and observe.

**Screenshots**

<img width="1792" alt="Screen Shot 2020-04-29 at 00 11 15" src="https://user-images.githubusercontent.com/4290500/80560182-033f0d00-89ae-11ea-9cf5-ee65e4de04e1.png">

**Environment (optional)**
 - Browser [e.g. chrome, safari] chrome 81
 - OS: [e.g. macOS] macOS

**Additional context (optional)**

PR #486, opened 2020-05-01
Canvas hotfix

mt-xing:
### Summary 

Per Matthew's request, Canvas tasks can now be deleted after unlinking Canvas. Also added better messaging.

### Test Plan 

Linking Canvas:
image

Unlinking Canvas:
image

Task view after unlink:
image


mt-xing:
@meganyin13 Does need to fix #483 before we can roll this out to production

PR #493, opened 2020-05-06
Seamless drag-and-drop experience

SamChou19815:
### Summary

The drag and drop user experience is far from being perfect.
To make order persistent, we have to save the updated order in the firestore. Although firestore seems to have ['latency compensation'](https://firebase.google.com/docs/firestore/query-data/listen#events-local-changes) that can avoid a roundtrip before updated data is pushed to remote, the listener update is still not **synchronous**, which causes a big trouble in dnd that relies an a synchronous state change.

In the past, we introduced `localTasks` as a cache in focus view to solve the problem. We are able to mostly avoid the dropped task bouncing back to the original position, but there are still some occasional flashes. In future view, such caching mechanism has not been implemented, so the resulting experience is even worse.

Recently, I encountered a similar problem in my side projects, and I found a more principled and easier way to solve it. Surprisingly, this easier way also makes the UX better. An crucial observation is that

> Redux store update is synchronous.

Therefore, instead of `setState` locally in a component, why don't we do similar stuff to the redux store? Then we can avoid creating two versions of truth. It's also easier to implement, since the functions we call to update on firebase already works on change sets, and our redux store updates also work on change set.

After this change, we are able to completely remove those ugly local caches and make the app more responsive!

### Test Plan

Try to drag and drop within focus view, from focus view completed to uncompleted, within a day in future view and between days.
No more flashes!

### DIFF
--- a/frontend/src/firebase/actions.ts
+++ b/frontend/src/firebase/actions.ts
@@ -20,6 +20,7 @@ import {
 } from 'common/lib/types/firestore-types';
 import { WriteBatch } from 'common/lib/firebase/database';
 import Actions from 'common/lib/firebase/common-actions';
+import { TaskWithChildrenId } from 'common/lib/types/action-types';
 import {
   reportAddTagEvent,
   reportEditTagEvent,
@@ -37,6 +38,7 @@ import { getAppUser } from './auth-util';
 import { db, database } from './db';
 import { getNewTaskId } from './id-provider';
 import { store } from '../store/store';
+import { patchTags, patchTasks, patchSubTasks } from '../store/actions';
 import { Diff } from '../components/Util/TaskEditors/TaskEditor/task-diff-reducer';
 
 const actions = new Actions(() => getAppUser().email, database);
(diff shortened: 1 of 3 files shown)


mt-xing:
Does this mean that we now have React components directly pushing data to the redux store?

SamChou19815:
> Does this mean that we now have React components directly pushing data to the redux store?

Yes. We first push to redux store, and push to firestore and we ensure that the later push to firestore will result in an unchanged state. This double change is far from ideal, but at least it's cleaner than a local cache.

SamChou19815:
Previously, we have a somewhat implicit invariant that firestore is the single source of truth. 
In fact, to be more accurate, the above statement should be split in half:

- W property: Firestore is the single source of truth for writes
- R property: Redux store is the single source of truth for reads

Here is a table that argues that the new approach is strictly better:

|                | Violates R property | Violates W property |
|----------------|------------|------------|
| read and write from local cache    | T          | T          |
| read and write from redux store | F          | T          |

The new approach still violates a desirable property, but violating only one is better than violating two.

frontend/src/firebase/actions.ts
```diff
@@ -425,6 +431,30 @@ export function completeTaskInFocus<T extends { readonly id: string; readonly or
  * @return a new list with updated orders.
  */
 export function applyReorder(orderFor: 'tags' | 'tasks', reorderMap: Map<string, number>): void {
+  const { tags, tasks } = store.getState();
+  if (orderFor === 'tags') {
+    const editedTags: Tag[] = [];
+    Array.from(reorderMap.entries()).forEach(([id, order]) => {
+      const existingTag = tags.get(id);
+      if (existingTag != null) {
```
ptwu, line 439: Since `.get()` in the tags map can only return the value or `undefined` at id in the function call above this line, can't this be refactored this be `existingTag !== undefined` instead?

SamChou19815: Changed

PR #495, opened 2020-05-08
Firebase Groups

meganyin13:
### Summary 

This pull request sets up the firebase functions for the `samwise-groups` collection. I have functionality for creating, join, leaving, and fetching groups as well as getting members in a group. Currently looking for feedback if this is on the right track 🤗 ~~because tbh the type-checker just told me what to write.~~

- [x] join group 
- [x] create group
- [x] leave group
- [x] fetch groups
- [x] unmount listeners functions
- [x] get members

Does any of this need to be connected to Redux? Or can I call these functions directly?

Once this is done, I need to connect it to the group tasks frontend (willing to work on this after last meeting, if necessary).

### Test Plan 

👀 Hope nothing breaks. 

### Notes 

Added `classCode` as an optional field for group because from Figma designs looks like users can only create groups associated with a class they have on Samwise:

image


common/src/types/store-types.ts
```diff
@@ -100,6 +100,17 @@ export type BannerMessageStatus = {
   readonly [I in BannerMessageIds]?: boolean;
 };
 
+/**
+ * The type of a group entry.
+ */
+
+export type Group = {
+  readonly name: string;
+  readonly members: string[];
```
SamChou19815, line 109: Use `readonly string[]` instead

frontend/src/firebase/actions.ts
```diff
@@ -366,7 +367,68 @@ export const removeSubTask = (
...
+};
+
+export const getMembers = async (
+  groupId: string,
+): Promise<string[]> => {
+  const groupDoc = await database.groupsCollection().doc(groupId);
+  const groupSnapshot = await groupDoc.get();
+  const group = await groupSnapshot.data() as FirestoreGroup;
+  return group.members;
+};
```
SamChou19815, line 427: Not sure why do we need to read this from firestore. If the listeners are correct, it should already be in redux store

frontend/src/firebase/actions.ts
```diff
@@ -366,7 +367,68 @@ export const removeSubTask = (
...
 /*
  * --------------------------------------------------------------------------------
- * Section 3: Other Compound Actions
+ * Section 3: Groups Actions
+ * --------------------------------------------------------------------------------
+ */
+
+export const joinGroup = async (
+  groupId: string,
+): Promise<void> => {
```
mt-xing, line 376: You're probably going to need to take one more argument here: an invite ID.
This way, we can easily query the list of invites via our security rules to ensure the user has permission to join that particular group.

firestore.rules
```diff
@@ -38,5 +38,9 @@ service cloud.firestore {
       allow create: if request.auth.token.email in get(/databases/$(database)/documents/samwise-groups/$(request.resource.data.group)).data.members 
       allow read, delete: if request.auth.token.email == resource.data.invitee
     }
+    match /samwise-groups/{group} {
+      allow create: if canCreate()
+      allow read, delete: if request.auth.token.email in get(/databases/$(database)/documents/samwise-groups/$(request.resource.data.group)).data.members
```
SamChou19815, line 43: Use `resource` instead of `request.resource`
- `request.resource` is your edits (i.e. the stuff you pass in `xxxCollection.doc(...).update()`)
- `resource` is the stuff already in the db.

frontend/src/firebase/listeners.ts
```diff
@@ -240,6 +250,17 @@ export default (onFirstFetched: () => void): (() => void) => {
     reportFirstFetchedIfAllFetched();
   });
 
+  const unmountGroupsListener = listenGroupChange(ownerEmail, (snapshot) => {
+    const data = snapshot.empty ? undefined : snapshot.docs;
```
SamChou19815, line 254: Is snapshot is empty, would `snapshot.docs` be an empty array. Now sure why do you special case here?

frontend/src/firebase/actions.ts
```diff
@@ -435,7 +435,11 @@ export const leaveGroup = async (
   const { email } = getAppUser();
   const newMembers: string[] = members.filter((m: string) => m !== email);
   const groupDoc = await database.groupsCollection().doc(groupID);
-  groupDoc.update({ members: newMembers });
+  if (newMembers.length === 0) {
+    groupDoc.delete();
+  } else {
+    groupDoc.update({ members: newMembers });
+  }
```
SamChou19815: await on those updates

SamChou19815:
Screen Shot 2020-05-09 at 14 12 05
😢

frontend/src/firebase/actions.ts
```diff
@@ -373,6 +374,74 @@ export const removeSubTask = (
...
+    return;
+  }
+  const members = groups.get(groupId)?.members;
+  const { email } = getAppUser();
+  // Check if user is already in the group
+  if (members === undefined || members.includes(email)) {
+    return;
+  }
+  groupDoc.update({ members: [...members, email] });
+};
```
mt-xing, line 404: I don't think we're going to be able to directly query Firestore for this operation. There's no real way for us to write the security rules that will allow us to pass in the invite ID to check against without writing it to the document, which we don't want. We might need to use a Cloud Function for this. We can talk about it at the meeting today.

SamChou19815: We might still be able to do this, but rely on a weaker security model, with the assumption that groupId is unguessable. Then invitations' only purpose is to tell you a specific groupId.

SamChou19815: Here is an idea that might work:
Store invitation and group in the same document with format like:
```js
{
  invitations: {
    // invited         -> invitee name
    'foo@bar.com': 'Beep Boop'
  },
  members: ['1@a.com', '2@b.com']
}
```

If 'foo@bar.com' accepts the invitation, it will now become:

```js
{
  invitations: { },
  members: ['1@a.com', '2@b.com', 'foo@bar.com']
}
```

The security rule can then do some map and set diff to validate this.
https://firebase.google.com/docs/reference/rules/rules.Map
https://firebase.google.com/docs/reference/rules/rules.Set
https://firebase.google.com/docs/reference/rules/rules.MapDiff

PR #499, opened 2020-05-09
Firestore Support for Sending Group Invitations

mt-xing:
### Summary 

Create firebase functions and add support in the Redux Store for the `samwise-group-pending-invites` collection, used to store invitations for people to join a group. 
There is now a function called `sendInvite()` that can be called to invite a user to a group.
Invitations are pulled down to the redux store of the receiving user.
Also changed Firestore security rules to match.

### Test Plan 

Manually added a document to Firestore, and then added some test code to print out `JSON.stringify(pendingInvites)` from the Redux store, and it shows up correctly:
image

Likewise, manually calling the function to send an invite works.
image
As does manually calling the function to reject an invite, though I'm not sure how to screenshot that.

### DIFF
--- a/frontend/src/firebase/actions.ts
+++ b/frontend/src/firebase/actions.ts
@@ -368,7 +368,18 @@ export const removeSubTask = (
 
 /*
  * --------------------------------------------------------------------------------
- * Section 3: Other Compound Actions
+ * Section 3: Groups Actions
+ * --------------------------------------------------------------------------------
+ */
+
+export const rejectInvite = async (inviteID: string): Promise<void> => {
+  const invitations = await database.pendingInvitesCollection().doc(inviteID);
+  invitations.delete();
+};
+
+/*
+ * --------------------------------------------------------------------------------
+ * Section 4: Other Compound Actions
  * --------------------------------------------------------------------------------
  */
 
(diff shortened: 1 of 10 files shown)


frontend/src/firebase/actions.ts
```diff
@@ -368,7 +369,36 @@ export const removeSubTask = (
...
+    group: groupID,
+    inviterName: userName,
+    invitee,
+  };
+  database.pendingInvitesCollection().doc().set(newInvitation);
+};
+
+export const rejectInvite = async (inviteID: string): Promise<void> => {
+  const invitations = await database.pendingInvitesCollection().doc(inviteID);
+  invitations.delete();
```
SamChou19815, line 396: I think it should be:

```suggestion
  await database.pendingInvitesCollection().doc(inviteID).delete();
```

You don't need to await on a document reference.

frontend/src/firebase/actions.ts
```diff
@@ -368,7 +369,35 @@ export const removeSubTask = (
...
+ */
+export const sendInvite = async (
+  groupID: string, userName: string, invitee: string,
+): Promise<void> => {
+  const newInvitation: FirestorePendingGroupInvite = {
+    group: groupID,
+    inviterName: userName,
+    invitee,
+  };
+  database.pendingInvitesCollection().add(newInvitation);
```
SamChou19815, line 391: add await here

mt-xing: Oops, lol

PR #500, opened 2020-05-09
View and Reject Invitations

mt-xing:
### Summary 

Implement basic UI to allow users to see and reject group invitations.

### Test Plan 

image

Can gracefully handle multiple invitations:
image

Disappears when all are dismissed:
image

Rejecting an invitation works.
First, confirm that an invitation exists in Firestore:
image

Then click reject:
image

Observe the invitation is gone:
image


mt-xing:
Happy 500th PR, everybody!

PR #503, opened 2020-05-09
Fix Regressions

### LINKED ISSUE #498: Future View Height Incorrect
**Describe the bug**
Future view height is broken at small screen heights.
This seems like an incorrect computation based on some vh, incorrectly accounting for the fixed height portions of the page.

**To Reproduce**
Open Samwise, either production or staging, and make your window shorter.

**Screenshots**
image

**Environment (optional)**
 - Browser: Chrome
 - OS: Windows 10

meganyin13:
### Summary 

Fixes #498 which was caused by setting `TaskCreator ` a blanket `height: 5vh` for large screen widths to fix the bottom margin. However, for low screen heights we should just allow `height: auto`. (There may be a shift at height 650px though)

Also fixes greyed out settings page title:
image

### Test Plan 
👀

### DIFF
--- a/frontend/src/components/GroupView/RightView/index.module.css
+++ b/frontend/src/components/GroupView/RightView/index.module.css
@@ -20,7 +20,7 @@
   overflow-y: auto;
 }
 
-h2 {
+.RightView h2 {
   display: inline-block;
   color: #808080;
   font-size: 26px;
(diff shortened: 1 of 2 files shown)


SamChou19815:
I can confirm that the settings text is fixed. However, I'm not sure whether the height fix fixes anything. I can see that when I gradually reduce the height of the screen, the bottom margin of future view keeps increasing.

PR #509, opened 2020-08-23
Update FA20 courses without exams

SamChou19815:
### Summary 

We shouldn't let Cornell's incompetence in providing dates impact Samwise usage.

Let's at least update the course info now.

### Test Plan 

CI


mt-xing:
Doesn't the final course roster come out tomorrow? Do we want to wait for that before importing the course list?

SamChou19815:
> Doesn't the final course roster come out tomorrow? Do we want to wait for that before importing the course list?

We still need to update the code to fetch FA20 info.

PR #513, opened 2020-08-23
Adopt prettier

SamChou19815:
### Summary 

The airbnb linter config introduces a lot of opinionated style lint rules. It helps us to establish a team wide style guide without fighting over styles, but it also adds significant burden to developers.

For example, a lot of style violations are not autofixable, so devs have to spend extra time to make linter happy. Some rules are even in conflict with each other sometimes, and devs need to find smart ways to work around them. What's worse, even after all these burdens, it's still not guaranteed that each AST corresponds to a unique style-guide approved output, so there can be still fighting over styles.

Therefore, I think it's better to introduce prettier for this case. Prettier formats all the code in an opinionated way, and it can be run in a pre-commit hook. Therefore, even if you code style looks like garbage, the pre-commit hook will fix it automatically for you like:

Screen Shot 2020-08-23 at 18 16 45

The diff is big, but almost all the changes are generated by `yarn format`.

### Test Plan 

CI, and use you 👀  to decide whether prettier produces nice output.


SamChou19815:
> Sg. Do you have a particular `prettierrc` file that you recommend we use? (other than just enable singlequote)

I put the config inside the package.json, so we don't need another dotfile in the root for

PR #516, opened 2020-09-02
Release to production for FA20

ptwu:
### Summary 

This release contains:

- [x] Course data for FA20
- [x] New TaskCreator UI
- [x] Seamless drag and drop
- [x] Misc accessibility improvements (button instead of span tags)

### Test Plan 

Play with [staging](https://samwise-dev.firebaseapp.com/).

- FA20 courses should be able to be added as tags ~~exams should show up~~ 
- Drag and drop should be more seamless (i.e. no flicker)
- New TaskCreator

### Notes 

Sam is still working on an involved migration to `next.js`, which may cause breaking changes.


SamChou19815:
> exams should show up

This is false 😢 . See https://registrar.cornell.edu/exams/fall-final-exam-schedule

ptwu:
> > exams should show up
> 
> This is false 😢 . See https://registrar.cornell.edu/exams/fall-final-exam-schedule

oops, I forgot this is not your normal semester. I'll fix the PR description.

SamChou19815 (dismissed):

meganyin13 (dismissed):
I was able to find my classes and in repeating date selection the new end dates do show up. Just a couple (not push-blocking) concerns:

> * New TaskCreator UI

This may be a disruptive change for users?

This issue and #498 still persists:
image

But since things are behind feature flag, we can look into it later.

mt-xing (dismissed):
LGTM, but the class adder dialog looks a little wonky on Windows:

image

As does the calendar:

image

Not sure if that's intentional, but it's also not push-blocking.

ptwu:
> LGTM, but the class adder dialog looks a little wonky on Windows:
> 
> image
> 
> As does the calendar:
> 
> image
> 
> Not sure if that's intentional, but it's also not push-blocking.

This might be because we changed these options to button tags. I can reproduce this as well on my Windows laptop.

ptwu:
just found some more regressions that are associated with the calendar. Not really sure if this is caused by changing span tags to buttons, since I don't think we did that for the calendar. In any case, I'm fixing the CSS to mimic what's in production right now.

SamChou19815 (approved):

jason-tung (approved):

PR #518, opened 2020-09-02
Migrate frontend toolchain to next.js

SamChou19815:
### Summary 

I'm very unhappy about CRA's slow evolution progress. It seems to be very close to a 4.0 release but never made it for months. See https://github.com/facebook/create-react-app/discussions/9131

I'm relatively sure that this kind of slow progress will continue even if by magic the CRA maintenance team suddenly puts more effort into it. The underlying reason is that CRA is trying to do too many things at once. In addition to a nice webpack config, it also bundles eslint and jest. Therefore, usually it needs to wait until both eslint and jest to land support for new TS features before it can cut a new release for TS.

Next.js is less ambitious and only takes care of the webpack stuff. It also allows us to customize a lot of stuff so that we don't need to wait on anything. It also has server-side-rendering support to allow samwise to get better SEO. It also supports [fast refresh](https://twitter.com/dan_abramov/status/1144630845103362048?lang=en) to improve dev experience.

This diff aims to get the initial migration done without starting to taking advantage of anything new offered by Next.js. I will document the changes in the breaking changes section.

### Test Plan 

Check whether everything works. I'll wait until I got 3 approving reviews to merge to build up this migration confidence.

### Notes

**DO NOT MERGE BEFORE #516 IS MERGED!**
**DO NOT MERGE BEFORE #516 IS MERGED!**
**DO NOT MERGE BEFORE #516 IS MERGED!**
**DO NOT MERGE BEFORE #516 IS MERGED!**
**DO NOT MERGE BEFORE #516 IS MERGED!**

### Breaking Changes

- Frontend entry point is now `src/pages/index.tsx`
- You can no longer import image assets. Instead, please put it inside `public` folder, hardcode its path in `assets/assets-constants.ts` with PUBLIC_PATH prepended, and import the constant.
- Now some code can be executed in SSR environment. To gate browser only code, use `process.browser`.
- Root of the app is now `<div id="__next" />` instead of `<div id="__root" />`
- We now maintain our own jest config since CRA is gone.
- Absolute imports have been mostly removed since the new tests don't like that.


SamChou19815:
@cornell-dti/samwise please try it out!

SamChou19815:
> Ok these are nits but I'm also seeing the pin icon next to add class shifting up and the future view arrows are a lighter color.
> 
> side-by-side

Good catch. These particular problems have been fixed now. We need more eyes on this migration to ensure nothing goes wrong. Right now a lot of stuffs might be slightly off probably because:

1. We completely changed how we render svg. Especially, all the css that targets svg elements will no longer work.
2. Next.js might have slightly different rule of ordering imported css, which might cause cascading behavior to be slightly different.

meganyin13 (commented):
I completely forgot to check the group view. Here are some things I noticed:

1. The personal view icon on the sidebar is very small. I think I set the height and width on the svg directly [here](https://github.com/cornell-dti/samwise/blob/11d04b5a46f75def5ad71ac2160da3f9dd0c1181/frontend/src/components/SideBar/index.module.css#L11) 
2. I think we could use those `&nbsp;` around the assignTo icon as well in the new task creator in GroupView
3. There was some more spacing around the color dot in the +New Group dropdown when you hover over it. Here is side by side (right is staging currently)


These are all the regressions I found. Maybe @ptwu could also take a look 👀

SamChou19815:
> I completely forgot to check the group view. Here are some things I noticed:
> 
> 1. The personal view icon on the sidebar is very small. I think I set the height and width on the svg directly [here](https://github.com/cornell-dti/samwise/blob/11d04b5a46f75def5ad71ac2160da3f9dd0c1181/frontend/src/components/SideBar/index.module.css#L11)
> 2. I think we could use those `&nbsp;` around the assignTo icon as well in the new task creator in GroupView
> 3. There was some more spacing around the color dot in the +New Group dropdown when you hover over it. Here is side by side (right is staging currently)
> 
> 
> 
> These are all the regressions I found. Maybe @ptwu could also take a look 👀

1, 2, 3 should all be fixed now.

meganyin13 (approved):
:dev-sam: refactor LGTM 🚀 

CRA-vs-nextjs

SamChou19815:
@ptwu @jason-tung @pihucode @mt-xing takes a look to see whether there are any other broken stuff.

ptwu (approved):
LGTM, played with everything I could think of and compared to prod.

jason-tung (approved):
I did notice a bug where changing the tag of a repeating task will make the popup shift down, but this does not seem to be tied with this PR in particular as the same bug can be reproduced on dev.
https://youtu.be/BbTn1V3c5_I

LGTM 👍

PR #519, opened 2020-09-02
Fix calendar styles to avoid regression

ptwu:
### Summary 

This PR fixes some style issues with our calendar date picker powered by `react-calendar`. Since the last prod release, the UI of the calendar picker has become messed up (but the functionality is still there). ~~We haven't bumped the version of `react-calendar` since,~~ (actually we have bumped react-calendar) so this is likely to have been that or due to the accessibility changes earlier in which we changed tags for buttons to `<button>`.

This issue was pointed out by @mt-xing in #516. 

### Test Plan 
**Before:**
fVX368D

**After**
IV3vBXD

Play with the calendar / add date feature on staging for this PR and compare to current prod (https://samwise.today/)

### Notes
Only tested so far on my Windows 10 laptop running on Chrome.

### DIFF
--- a/frontend/src/components/TitleBar/Tags/TagAdder.module.css
+++ b/frontend/src/components/TitleBar/Tags/TagAdder.module.css
@@ -31,6 +31,7 @@ input.SearchInput:focus {
 }
 
 button.DropdownItem {
+  border: none;
   display: block;
   width: 100%;
   font-size: 14px;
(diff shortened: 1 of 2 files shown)


meganyin13 (approved):
LGTM. I think this might be because we bumped react-calendar in #511.

PR #525, opened 2020-09-07
Convert all css to scss

SamChou19815:
### Summary 

Convert all css to scss to use more powerful feature like extend, mixin, operators, etc. Hopefully we can use these more powerful features to replace ugly hacks in the future.

### Test Plan 

No real code change, so everything should still work.

Screen Shot 2020-09-07 at 10 52 24
Screen Shot 2020-09-07 at 10 52 43


meganyin13 (approved):
No-CSS

PR #527, opened 2020-09-09
Avoid compiling TS code in common package

SamChou19815:
### Summary

After we switched to next.js for frontend, we can easily customize the webpack behavior. We already added a config that lets webpack calls babel on common package, so we don't need to compile the TS code to get JS for frontend.

Similarly, ncc can work on TS sources directly, so functions package also doesn't need a compiled common package.

Therefore, let's stop producing compiled JS in common package.

### Test Plan

Check staging site still works.

### Notes

After this PR is merged, you should remove `common/lib` folder since the compiled code there is no longer useful.


meganyin13 (approved):
I can't find a good meme but that shouldn't block review 😆

PR #531, opened 2020-09-11
Adopt tssa to statically analyze impact of each PR

SamChou19815:
### Summary

In the past, we had plenty of SEVs that is caused by some innocent looking changes that greatly messed up code elsewhere. In this diff, I propose to use `@dev-sam/tssa` to statically analyze each diff to report all files that can be potentially affected by your diffs.

Right now, tssa will perform the analysis quite naively:

1. Build a dependency graph of the entire project
2. In the CI, call `git diff` to get a list of changed files. TSSA group those changed file paths into workspaces.
3. For a TS file change, we compute its reverse dependencies and transitive reverse dependencies
4. For a css file change, we compute both its transitive forward dependencies and transitive reverse dependencies. Both directions are needed due to css' cascading nature.

Let's experiment on this to see whether it can help us catch problems.

### Test Plan

To let everyone see the result, I made some comment only changes in a TS file and in a css file. See PR comment made by bot.

### DIFF
--- a/frontend/src/components/UI/ModeIndicator.module.scss
+++ b/frontend/src/components/UI/ModeIndicator.module.scss
@@ -1,5 +1,6 @@
 .Indicator {
   position: fixed;
+  // TODO: move all z-index numbers into a css constant file to avoid increasingly giant numbers like this.
   z-index: 1000;
   background: #f44336;
   padding: 5px;
--- a/.github/workflows/tssa-analysis.yml
+++ b/.github/workflows/tssa-analysis.yml
@@ -1,0 +1,28 @@
+name: TSSA
+on: pull_request
+
+jobs:
+  analyze:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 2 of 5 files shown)


frontend/src/components/UI/ModeIndicator.module.scss
```diff
@@ -1,5 +1,6 @@
 .Indicator {
   position: fixed;
+  // TODO: move all z-index numbers into a css constant file to avoid increasingly giant numbers like this.
```
ptwu, line 3: Maybe we can file a GitHub issue for this? 👀🌾🚜

ptwu (approved):

frontend/src/components/UI/ModeIndicator.module.scss
SamChou19815: https://github.com/cornell-dti/samwise/issues/532

PR #538, opened 2020-09-18
Update course info check frequency

meganyin13:
### Summary 

This PR reduces the course info check frequency as suggested in #533. We no longer run course info check on push to master and instead run it every week before/at our meeting (Saturdays 3pm). 

### Test Plan 

👀 Action should no longer be triggered on push to master but it should run on Saturday 3pm EDT/19:00 UTC (if I got the cron syntax right).

image

### DIFF
--- a/.github/workflows/e2e-course-info.yml
+++ b/.github/workflows/e2e-course-info.yml
@@ -1,6 +1,6 @@
 on:
   schedule:
-    - cron: '0 15 * * 6'
+    - cron: '0 19 * * 6'
   pull_request:
     paths:
       - .github/workflows/e2e-course-info.yml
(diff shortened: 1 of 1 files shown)


SamChou19815 (approved):
It will be helpful if you attach the screenshot of the github editor in the test plan like

Screen Shot 2020-09-17 at 23 23 44

From the screenshot, we can see that it's not actually 3pm on Saturday **in ET**. However, it currently translates to 11am and translates to 10am without daylight saving. Anyway, this still satisfies the need for it to be run before the subteam meeting, so I'm going to accept this.

meganyin13:
wut there is an editor for this? ya I thought things maybe wouldn't already be in ET... will fix

PR #546, opened 2020-09-26
Flattern frontend/src/components/GroupView folder layout

SamChou19815:
### Summary

We have too many nesting in this folder and within each folder there is only one or two component with generic name like `index`, `Member`. This is not good for code search. For example, when you search `Member` in vscode to see how many components are using it, you can get too many irrelevant results.

This diff gives those components a longer name by prefix the generic name with all the containing folders, and then flatten the folder layout.

Previous:

```
GroupView/
  MiddleBar/
    People/
      index
      Member
    TaskQueue/
      index
  RightView/
    GroupTaskRow/
      index
      GroupTask
      GroupTasksContainer

```

Now:
```
GroupView/MiddleBar/GroupViewMiddleBarPeopleList
GroupView/MiddleBar/GroupViewMiddleBarMemberRow
GroupView/MiddleBar/GroupViewMiddleBarTaskQueue
GroupView/RightView/GroupTask
GroupView/RightView/GroupTasksContainer
GroupView/RightView/GroupTaskRow
```

### Test Plan

Preview works.


meganyin13 (approved):
Ok yup this is probably better 👍

PR #547, opened 2020-09-26
Auto update username when the user record already exists

SamChou19815:
### Summary

Looks like a bug when we only update photo but not username. @ptwu 

### Test Plan

Edit your name in firestore manually in db. Refresh the page, and check firestore record updates to your google username again.


ptwu:
I didn't update names for users in Firestore since we don't usually anticipate that names of students will change, but I guess this doesn't hurt.

ptwu (approved):

PR #548, opened 2020-09-26
Render GroupView and SideBar using groups from firestore

SamChou19815:
### Summary 

We already have redux store connected to the firestore collection. Let's read from redux store to replace hardcoded data.

### Test Plan 

It is correctly rendering my test group with these info:
Screen Shot 2020-09-26 at 16 49 35

Screen Shot 2020-09-26 at 16 49 18


meganyin13 (approved):
LGTM. The only thing is I kept `classCode` as optional in case we want to allow creating groups not associated with classes but it doesn't seem like from designs we will allow that so this should be good 👍

PR #549, opened 2020-09-26
Remove duplicate setting of env var in ci-frontend.yml

SamChou19815:
### Summary 

I might be sleepy when I did this.

### Test Plan 

CI


meganyin13 (approved):
:dev-sam-sleeps:

PR #537, opened 2020-09-18
Store user data in Firestore upon login

ptwu:
### Summary 

This pull request allows for population of the `samwise-users` collection in the Firestore. Every time a user logs in, the `useEffect` hook in `LoginBarrier` calls a newly created Firebase action that adds a new document to `samwise-users` with `name` and `photoURL` fields (this can be scaled to include more fields in the future, should we want to store more about a user). This will allow for searching names along with corresponding profile photos of group members associated with emails in Group View.

If the user's name is already stored in `samwise-users`, we don't make any change to the name, but update the profile photo.

### Test Plan 

dqv4hpR

PR #550, opened 2020-09-27
Render a list of user profiles in the group from firestore

SamChou19815:
### Summary 

This PR changes the member list from a hardcoded names to a list of profiles from firestore. I extract the logic of fetching users into a custom hook since the logic is a bit involved. See the inline comments for its flow.

To allow reading the profile from client side listeners, we do have to loosen the security rules a little bit. I currently loosen the rule to allow read if the user's email ends with @cornell.edu. I think this setup doesn't weaken any security measures. Currently it will only expose name and profile picture, which you can get by typing the email in cornell email composer anyways.

### Test Plan 

Screen Shot 2020-09-26 at 23 44 22

### DIFF
--- a/common/types/store-types.ts
+++ b/common/types/store-types.ts
@@ -104,13 +104,19 @@ export type BannerMessageStatus = {
 /**
  * The type of a group entry.
  */
-
 export type Group = {
   readonly id: string;
   readonly name: string;
   readonly members: readonly string[];
   readonly deadline: Date;
   readonly classCode: string;
+};
+
+/** The user profile of any samwise user. */
+export type SamwiseUserProfile = {
+  readonly email: string;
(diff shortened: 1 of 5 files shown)


meganyin13 (approved):
Looks good but I think you forgot @jason-tung 😐

SamChou19815:
> Looks good but I think you forgot @jason-tung 😐

@jason-tung has not played with samwise staging since #537 is merged, so we don't have his info in the `samwise-user` collection. If we include jason's email, it will crash. (Not a problem of code, since we will maintain the invariant that a email can only be added to a group if the person accepts the invitation, which would imply they signed in samwise once.)

That being said, @jason-tung should play with samwise staging more often 🤔

Issue #552, opened 2020-09-28
[maintenence] Replace Sam's script to fetch courses from course roster with TS code


SamChou19815:
## Motivation

Currently, Sam's script `git clone` a repo of kotlin code, builds it on the fly, and then use it to fetch courses from course roster. Sam builds this library as a general purpose one, and it would be best used if you are using kotlin and you actually need the precise types of those course json. For samwise's use case, this seems to be an overkill.

In addition to being an overkill, the setup has these overheads:

1. `git clone` and build kotlin code on the fly takes significant amount of time.
2. The library will output the entire json from the course roster, even though we only need a tiny subset of them. (We especially don't care about things like description and class time etc). We wasted a lot of time serializing and deserializing this json.

Also, it requires bash to hook everything together, which is not cross-platform. It creates additional burden for windows devs on the team.

## Guide

Devs interested in doing the task can reference the code Sam wrote for course plan: https://github.com/cornell-dti/course-plan/blob/master/src/requirements/fetcher.ts
Ideally, the should be some unit tests on the fetcher to ensure things work.

PR #554, opened 2020-10-01
Adjust prelim and final URLs and fix parser for FA20

### LINKED ISSUE #510: Make course-info script fetch FA20 final info when Cornell releases it

ptwu:
### Summary 

This pull request uncomments the prelim parsing code, as the prelim schedule has been posted on the registrar [here
](https://registrar.cornell.edu/exams/fall-prelim-exam-schedule). The format of the exam listings has also changed, and the HTML parser has been updated accordingly as explained below.

We can't add finals, as the registrar doesn't have this data yet.

- [x] fixes #510  (at least partially since we're uncertain if the registrar will provide finals data at all) 

While the spring prelim schedule from SP20 is no longer accessible, I used archive.org to find a cached version of the page. Previously, dates were in the format `MM/DD/YYYY` for each class, but now they are in the format `DAY_OF_WEEK DD MONTH_NAME`. The HTML parser has been updated accordingly to create a Date object that corresponds to the new date format.

SP20
image

FA20
image

### Test Plan 

Run the bash script in `/course-info` by running `./run` in the directory. See that no errors occur. You can cross-check the generated `.json` file to see if the `examTimes` field for a certain class object is defined. For example, CS 3110 has a prelim this coming Sunday, and its exam time has been populated.

image

### DIFF
--- a/course-info/src/fetch-exam.ts
+++ b/course-info/src/fetch-exam.ts
@@ -16,7 +16,7 @@ async function fetchExamText(url: string): Promise<string> {
 }
 
 function parsePrelimLine(line: string): ExamInfo {
-  const segments = line.split(' ');
+  const segments = line.split(/\s+/);
   const subject = segments[0];
   const courseNumber = segments[1];
   const dateTimeString = `${segments[3]} ${segments[4]} ${currentYear}`;
--- a/course-info/src/fetch-exam.ts
+++ b/course-info/src/fetch-exam.ts
@@ -20,9 +20,11 @@ function parsePrelimLine(line: string): ExamInfo {
   const subject = segments[0];
   const courseNumber = segments[1];
   const dateTimeString = `${segments[3]} ${segments[4]} ${currentYear}`;
+  const dateHours = parseInt(segments[5].substring(0, 2), 10);
+  const dateMinutes = parseInt(segments[5].substring(3, 5), 10);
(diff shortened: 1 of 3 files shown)


SamChou19815:
> [tssa]
> 
> No notable changes.

tssa doesn't support analyzing files not belonging to yarn workspace :(
I think it will be fixed once we fixed #552.

course-info/src/fetch-exam.ts
```diff
@@ -16,18 +16,11 @@ async function fetchExamText(url: string): Promise<string> {
...
-    if (s !== '') {
-      datetimeString = s;
-      break;
-    }
-  }
-  const date = new Date(datetimeString);
+  const dateTimeString = `${segments[3]} ${segments[4]} ${currentYear}`;
+  const date = new Date(dateTimeString);
   date.setFullYear(currentYear);
   date.setHours(19, 30); // 7:30 prelim time
```
SamChou19815, line 25: the screenshot shows some exams at 4:00PM, so I guess we should somehow change this?

ptwu: Done.

SamChou19815 (approved):
ashipshipping-shipshippingshipping-ships-a-ship-shipping-ship-shipping-shipping-ships

PR #558, opened 2020-10-03
Add search bar to assign group members

pihucode:
### Summary
Replaces the tag picker in normal view with a group member picker when in group view. Group members can be searched by either their names or their Cornell netIDs.

- [x] replace tag picker with a group member picker

### Test Plan
group member picker

### DIFF
--- a/common/types/store-types.ts
+++ b/common/types/store-types.ts
@@ -112,6 +112,14 @@ export type Group = {
   readonly classCode: string;
 };
 
+/**
+ * The type of a group member.
+ */
+export type GroupMember = {
+  readonly netId: string;
+  readonly name: string;
+};
+
 /** The user profile of any samwise user. */
 export type SamwiseUserProfile = {
   readonly email: string;
--- a/common/types/store-types.ts
+++ b/common/types/store-types.ts
@@ -112,14 +112,6 @@ export type Group = {
   readonly classCode: string;
 };
 
-/**
- * The type of a group member.
- */
-export type GroupMember = {
-  readonly netId: string;
-  readonly name: string;
-};
-
 /** The user profile of any samwise user. */
 export type SamwiseUserProfile = {
   readonly email: string;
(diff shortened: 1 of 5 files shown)


common/types/store-types.ts
```diff
@@ -112,6 +112,14 @@ export type Group = {
   readonly classCode: string;
 };
 
+/**
+ * The type of a group member.
+ */
+export type GroupMember = {
+  readonly netId: string;
+  readonly name: string;
```
ptwu, line 120: We might also want a field for photo URL

frontend/src/components/TaskCreator/GroupMemberPicker.tsx
```diff
@@ -67,8 +75,9 @@ function GroupMemberPicker({
     <div className={styles.Main}>
       {displayedNode(tag === NONE_TAG_ID)}
       {opened && (
-        <div className={styles.NewTaskClassPick}>
-          <TagListPicker onTagChange={onTagChange} />
+        // <div className={styles.NewTaskClassPick}>
```
ptwu, line 78: Should delete commented code

frontend/src/components/Util/GroupMemberListPicker/SearchGroupMember.tsx
```diff
@@ -0,0 +1,63 @@
...
+  let i = 0;
+  members.forEach((member: GroupMember) => {
+    const { netId, name } = member;
+    const id = netId;
+    memberOptions.push({
+      key: i,
+      value: name,
+      id,
+    });
+    i += 1;
```
ptwu, line 32: instead of keeping track of a mutable variable and incrementing it, we can use the second argument of `.forEach()` to get indices of `members`.

ptwu (approved):
good stuff! left a few comments / code quality fixes

common/types/store-types.ts
SamChou19815: It's probably good for now. This PR just use hardcoded data, and we will naturally switch to use the existing `SamwiseUserProfile` once we integrate with real data.

ptwu (approved):

PR #563, opened 2020-10-03
Fix group tasks icon in personal view

meganyin13:
### Summary

Originally, we have the default icon next to the task title be the repeating icon, so now when we add a new type `GROUP`, it will default to the repeating icon. Let's make the Placeholder (empty) the default and if it is `ONE_TIME` and not a canvas task, we can show the `DragIcon`. All other cases, leave the icon blank.

### Test Plan

Make group, one time, and repeating tasks and check the icons. 

icons


jason-tung (commented):
lgtm on staging but I am curious if it is intended that the else case would be no icon at all when the tasks are indeed draggable.

image


the draggable tasks w/o icon are "GROUP" type

SamChou19815 (approved):
Approve to unblock unless people have strong opinion on this.

PR #565, opened 2020-10-06
Adopt tssa 0.0.9

SamChou19815:
### Summary 

The new tssa can support function-level dependency analysis, so it can produce a more precise list of symbols a change can affect.
It also contains @meganyin13's change to break down css change affect list by file.

### Test Plan 

Test when merged into master.


meganyin13 (approved):
Let's see if this works 😅
shipitparrot

PR #562, opened 2020-10-03
September Release

SamChou19815:
### Summary 

Release everything we did in September and first 3 days of October into prod.

Notable changes:

- [ui] @IshikaJain fixed accessibility issues for date, tag, and focus picker
- [group-task, ui] @meganyin13 removed the incorrect repeating task icon from group task in future view.
- [group-task] Several updates to group task by (@SamChou19815, @pihucode, @ptwu). They are gated so just check site without group tasks still work.
- [chore] Updated course json multiple times and the update frequency is reduced by @meganyin13. Now it has prelim exams thanks to @ptwu .
- [infra] Migrated frontend toolchain from CRA to next.js

### Test Plan 

Staging

PR #566, opened 2020-10-07
Make class tags center aligned in TagPicker

ptwu:
### Summary 

This pull request is a 1 line fix that just makes new task class tags under tag picker have center alignment, fixing the issue pointed out by @meganyin13 in #562.

### Test Plan 

image

play with tagpicker :eyes:


SamChou19815 (approved):
1line

Issue #567, opened 2020-10-07
Assign group task does not work


jason-tung:
**Describe the bug**
Creating group task does not create a task with the proper group field. It creates a doc in Firestore with the group field as the name rather than the id (which should be the docid of the group in the samwise-groups collection).

The owner is also appears to default to the person who makes it and is thus not assignable to others.

**To Reproduce**
Try adding any group task and observe Firestore.

**Screenshots**
![image](https://user-images.githubusercontent.com/33106747/95371303-036dfd80-08a8-11eb-8db1-e191875f661c.png)

jason-tung:
duplicate of #561

PR #568, opened 2020-10-07
GroupTasks pulling real data from Firestore

### LINKED ISSUE #560: [group-task] Fetch and render group tasks in group view
- [ ] Listen to firestore on all the group tasks (not only the tasks where you are the owner, but also tasks with group id among the groups you are in)
- [ ] Partition these group tasks according to owners
- [ ] Render these group tasks into the correct row in group view

jason-tung:
### Summary 

Replace temporary data pulled from focus view with actual data from Firestore according to the group and user.

resolves #560 

### Test Plan 

Play with it on staging. 
Note: the add tasks is actually broken because it adds based on group name rather than the group id and only assigns the task to yourself. See: #567

The way I have been adding tasks is just using the add task feature at the top and editing fields manually in Firestore.
image

### Notes 

Subtasks are not yet rendered/supported until there is some util for converting the string[] from Firestore type to the Store type SubTask[]

### DIFF
--- a/frontend/src/components/GroupView/RightView/GroupTaskRow.tsx
+++ b/frontend/src/components/GroupView/RightView/GroupTaskRow.tsx
@@ -7,7 +7,7 @@ import styles from './GroupTaskRow.module.scss';
 
 type Props = {
   memberName: string;
-  readonly tasks: Task[];
+  readonly tasks: readonly Task[];
 };
 
 const clickAddGroupTask = (e: SyntheticEvent<HTMLElement>): void => {
--- a/frontend/src/components/GroupView/RightView/GroupTaskRow.tsx
+++ b/frontend/src/components/GroupView/RightView/GroupTaskRow.tsx
@@ -1,11 +1,13 @@
 import React, { ReactElement, SyntheticEvent, KeyboardEvent } from 'react';
 import { faPlus } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { Task } from 'common/types/store-types';
 import GroupTasksContainer from './GroupTasksContainer';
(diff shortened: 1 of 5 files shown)


frontend/src/components/GroupView/RightView/GroupTaskRow.tsx
```diff
@@ -1,11 +1,13 @@
 import React, { ReactElement, SyntheticEvent, KeyboardEvent } from 'react';
 import { faPlus } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { Task } from 'common/types/store-types';
 import GroupTasksContainer from './GroupTasksContainer';
 import styles from './GroupTaskRow.module.scss';
 
 type Props = {
   memberName: string;
+  readonly tasks: Task[];
```
SamChou19815, line 10: nits, add `readonly` before `Task[]` to make the array is readonly.

frontend/src/components/GroupView/index.tsx
```diff
@@ -44,12 +50,35 @@ const useGroupMemberProfiles = (
...
+        s.forEach((it) => {
+          const d = it.data();
+          const task = {
+            ...d,
+            id: it.id,
+            metadata: { type: 'GROUP', date: d.date.toDate(), group: d.group } as GroupTaskMetadata,
+            children: [] as readonly SubTask[],
+          } as Task;
+          // get rid of groupTaskArray from dependency array
+          setGroupTaskArray((ary) => [...ary, task]);
```
SamChou19815, line 71: nits: do not use single-letter variables. Use full names whenever possible.

SamChou19815 (approved):
Some small nits, but overall looks good. The subtask problem would better to be dealt with in another diff, since it might involve some big refactors.

PR #571, opened 2020-10-09
Connect group tasks UI to backend for creating/leaving and profile images

### LINKED ISSUE #559: [group-task] Connect real data to component currently using hardcoded data
- [x] Render a list of users name with profile picture within a group (bonus: create a component for profile pic so it can be reused)
- [x] connect leave group to UI
- [x] connect create group to UI

ptwu:
### Summary 

This pull request implements group creation, group leaving for a user's groups, and rendering profile pictures for group members in the UI. 

We create a group using the `createGroup` Firebase action in `actions.ts`, which correspond to each of the class ID picker buttons when hovering on the plus button on the left view switcher bar (this has already been implemented, but with no functionality). The default values chosen are a group `name` of `New group`, `deadline` of the current day, and `member` a singleton array of just the user who created it. For group leaving, we use the `leaveGroup` action, which is also fairly straightforward. However, we also want the page to rerender so that the group that we left is no longer displayed on the user's screen, so we need to lift state up by passing the `changeView` callback from `ViewSwitcher` down to the `GroupViewMiddleBarPeopleList`. Lastly, profile pictures are rendered using the data attached to each `SamwiseUserProfile` type, but a custom `ProfileImage` component has been defined so we can reuse it. Currently, the `ProfileImage` component is just a rounded image with no specified dimensions.

- [x] fixes #559 
- [x] implements profile picture rendering with custom component
- [x] implements group leaving and group creation functionality in the UI with respect to backend
- [x] accessibility improvement for class item picker when creating a group

### Test Plan 

GIF of how you can test the changes in the frontend:

gif

Correspondingly, you should see the side-effects of these changes in the Firestore.

### Notes 

For accessibility, I refactored the class ID picker for creating a group to be a `<button>` tag. This required overriding several default `<button>` styles, which is why I used `!important` in some of the CSS styles for the `ClassItem` class.

### DIFF
--- a/frontend/src/components/SideBar/AddGroupTags.module.scss
+++ b/frontend/src/components/SideBar/AddGroupTags.module.scss
@@ -7,8 +7,14 @@
 }
 
 .ClassItem {
+  background: none;
+  border-radius: 0 !important;
   display: flex;
-  padding: 0 0.5em;
+  font-size: 100%;
+  height: 34px !important;
+  margin: 0 !important;
+  padding-left: 10px;
+  width: 100% !important;
 }
 
 .ClassItem:first-of-type {
@@ -32,6 +38,7 @@
   border-radius: 50%;
   background-color: skyblue;
   position: relative;
+  margin-right: 8px;
   top: 5px;
 }
 
--- a/frontend/src/components/SideBar/AddGroupTags.tsx
+++ b/frontend/src/components/SideBar/AddGroupTags.tsx
@@ -1,14 +1,19 @@
-import React from 'react';
+import React, { ReactElement } from 'react';
 import { Tag } from 'common/types/store-types';
 import { store } from '../../store/store';
 import styles from './AddGroupTags.module.scss';
+import { createGroup } from '../../firebase/actions';
 
 type AddGroupTagsProps = {
   readonly show: boolean;
   readonly setShow: (show: boolean) => void;
 };
 
-export default function AddGroupTags({ show, setShow }: AddGroupTagsProps): React.ReactElement {
+function newGroup(classCode: string): void {
+  createGroup('default group name', new Date(), classCode);
+}
+
+export default function AddGroupTags({ show, setShow }: AddGroupTagsProps): ReactElement {
   const { tags } = store.getState();
   const classes: Tag[] = [];
   tags.forEach((t: Tag) => {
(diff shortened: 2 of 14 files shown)


frontend/src/components/SideBar/AddGroupTags.tsx
```diff
@@ -1,14 +1,19 @@
...
+import { createGroup } from '../../firebase/actions';
 
 type AddGroupTagsProps = {
   readonly show: boolean;
   readonly setShow: (show: boolean) => void;
 };
 
-export default function AddGroupTags({ show, setShow }: AddGroupTagsProps): React.ReactElement {
+function newGroup(classCode: string): void {
+  createGroup('New group', new Date(), classCode);
```
SamChou19815, line 13: Maybe we should use classCode as group name?

frontend/src/components/SideBar/AddGroupTags.tsx
```diff
@@ -25,15 +30,19 @@ export default function AddGroupTags({ show, setShow }: AddGroupTagsProps): Reac
...
+                  type="button"
+                  className={styles.ClassItem}
+                  onClick={() => newGroup(name.split(':')[0])}
+                >
+                  <div className={styles.Color} style={{ backgroundColor: color }}>
                     {' '}
                   </div>
-                  <p>{t.name.split(':')[0]}</p>
-                </div>
+                  <p>{name.split(':')[0]}</p>
```
SamChou19815, line 44: I'm not sure what is `name.split(':')[0]` doing. Maybe you can store it into a variable and give the variable a explanatory name?

frontend/src/components/SideBar/AddGroupTags.module.scss
```diff
@@ -7,8 +7,14 @@
...
 .ClassItem {
+  background: none;
+  border-radius: 0 !important;
   display: flex;
-  padding: 0 0.5em;
+  font-size: 100%;
+  height: 34px !important;
+  margin: 0 !important;
+  padding-left: 10px;
+  width: 100% !important;
```
SamChou19815, line 17: Is there a way we can avoid these `!important`? Maybe we can increase the css specificity for this. Not push blocking at this moment.

frontend/src/components/SideBar/AddGroupTags.tsx
meganyin13: I think this is getting the course department and number like get `CS 3410` from `CS 3410: Systems Programming` but yes new variable name would be more readable 👍

frontend/src/components/SideBar/AddGroupTags.tsx
meganyin13: I thought we would let them specify their own group name?

ptwu: I think Sam means to use the class code as the default name. What do you guys think about `'My CS 3110 Group'` or `New CS 3110 Group` (if it's for CS 3110)

frontend/src/components/SideBar/AddGroupTags.module.scss
ptwu: My rationale is in the `Notes` section, but to be fairly honest it would take me a nontrivial amount of time to try and increase the specificity. At the expense of this slightly suboptimal engineering decision, maybe it's best to move fast since we're sunsetting this sem.

frontend/src/components/SideBar/AddGroupTags.tsx
ptwu: [went with `New CS 3110 Group`](https://github.com/cornell-dti/samwise/pull/571/commits/bc9d22eadd457f85e3ee9ad7fbe98d6caef40264)

frontend/src/components/SideBar/AddGroupTags.tsx
ptwu: done

SamChou19815 (approved):
lgtm

PR #573, opened 2020-10-10
Deleted save button for group task

### LINKED ISSUE #555: [bug] Group task's editor should not have save button
## Problem

Right now group task has a save button, but there shouldn't be one. We have save button for repeating task because there is forking issue, but group task doesn't have forking issue.

Screen Shot 2020-10-01 at 15 06 45

## Guide

Intentionally omitted to make it as an exercise to onboard people to the codebase.

IshikaJain:
### Summary 
Deleted save button for group task but made sure repeating task still had the save button

- [x] Fixed #555 

### Test Plan 

Made sure that save button isn't there for group task but is there for repeating task
Screen Shot 2020-10-10 at 3 11 30 PM
Screen Shot 2020-10-10 at 3 12 09 PM


SamChou19815:
Already done in #574

Issue #561, opened 2020-10-03
[group-task] Assign people to group task


SamChou19815:
- [x] Finish searching group members feature with real data
- [x] When user click the assign button, task creator should pop-up with pre-filled assignee.

SamChou19815:
looks like it's done

PR #579, opened 2020-10-10
Assign member to group task

pihucode:
### Summary 

This pull request is the implements issue #561 

This pull request replaces the hard-coded user data used in searching with real data from `groupMemberProfiles `and renders `TaskCreator `with the corresponding member prefilled in the assigned member slot.

When the assign to button is clicked, a function is called to change the `taskCreatorOpened `and `assignedMember `states in `RightView `which will then be passed into the `TaskCreator`. State hooks will be used in `RightView `to pass down to the child class `TaskCreator `which will use `taskCreatorOpened `to determine whether to display the task creator. 

- [x] Finish searching group members feature with real data
- [x] When user click the assign button, task creator should pop-up with pre-filled assignee.
- [x] Update search bar for group members so that all members are displayed when search input is empty

### Test Plan 
Play around with the task creator.

### Notes 
One issue is that opening the `TaskCreator `through the "Assign new task" button does not cause `TaskCreator `to close upon clicking outside of the task creator. To ameliorate this, the `TaskCreator `can be opened and closed upon clicking on the "Assign new task" buttons.


frontend/src/components/GroupView/RightView/index.tsx
```diff
@@ -14,24 +26,42 @@ const EditGroupNameIcon = (): ReactElement => {
   return <SamwiseIcon iconName="pencil" className={styles.EditGroupNameIcon} onClick={handler} />;
 };
 
-const RightView = ({ group, groupMemberProfiles }: Props): ReactElement => (
-  <div className={styles.RightView}>
-    <div className={styles.GroupTaskCreator}>
-      <TaskCreator view="group" group={group.name} />
-    </div>
+export default class RightView extends React.PureComponent<Props, State> {
```
SamChou19815, line 29: Is there a reason why we have to use a class component instead of function component and with hooks

pihucode: It slipped my mind that React had a state hook, I can change `RightView` back to a functional class in a later commit.

frontend/src/components/GroupView/RightView/index.tsx
```diff
@@ -7,24 +7,51 @@ import TaskCreator from '../../TaskCreator';
...
+
 type Props = {
   readonly group: Group;
-  readonly groupMemberProfiles: readonly SamwiseUserProfile[];
+  readonly groupMemberProfiles: SamwiseUserProfile[];
   readonly tasks: readonly Task[];
 };
 
 const RightView = ({ group, groupMemberProfiles, tasks }: Props): ReactElement => {
+  const [{ taskCreatorOpened, assignedMember }, setState] = useState<State>({
```
ptwu, line 22: Maybe make a more descriptive name for this hook instead of `setState`, and even better maybe we can have a type for this data so we can pass a parameter into the generic here.

ptwu (commented):
image

Might also be good to let users pick from the dropdown by just clicking anywhere within each element of the list, as a user that's what I expected. Also a `cursor: pointer` attribute will help with UX there.

image

Here, I noticed the assigned to button after assigning someone has the `X` pushed off the button a bit. Is it possible to make the button fit both the close button and the name?

Otherwise, lgtm!!

SamChou19815 (approved):
approve to unblock

PR #582, opened 2020-10-13
Deploy Firestore rules

meganyin13:
### Summary
This PR adds deploying firestore rules as part of the CD process when we merge into staging/release. We want the [firestore.rules](https://github.com/cornell-dti/samwise/blob/master/firestore.rules) to be the single source of truth for security rules. Hopefully, this can prevent SEVs.

### Test Plan
I think we have a dummy rule for `demo-comment-board` that I deleted. If this works, on CD it should trigger a new firestore rules deploy which we can see on the dashboard.

### Notes
I got permission to steal this task from @jason-tung


SamChou19815 (approved):
LGTM.

Nit: Can we copy down the current rules in staging project so that the latest changes won't get overridden?

PR #586, opened 2020-10-15
Move ModalOptions css to eliminate 4 no-descending-specificity warnings

SamChou19815:
### Summary 

Move the css around so that we have 4 less css warnings.

### Test Plan 

dti-github bot says that there is no significant lines, which means that I just moved around some stuff 👀


meganyin13 (approved):
maybe a good idea for farms 👀

PR #587, opened 2020-10-15
Fix autosave functionality for group task

SamChou19815:
### Summary 

#574 only removed the save button for group task but it didn't fix the logic for autosave. This diff fixes that.

### Test Plan 

Broken on master:
broken

Fixed in this diff:
fixed


ptwu (approved):
100% original meme
image

PR #589, opened 2020-10-15
Avoid css child selector for Tag and Date Display component

SamChou19815:
### Summary 

This usage of css child selector is particularly bad. It's actually acting on the `SamwiseIcon`'s `span` container. It's bad because it assumes the implementation detail of its child component, which makes components more tightly coupled.

This diff makes `SamwiseIcon`'s span container css class overridable, so we no longer need this hack.

### Test Plan 

master:
master

this diff:
this

### Notes 

Q: Who created this bad css code?
A: It's me in #518. 😅


ptwu (approved):
LGTM

PR #590, opened 2020-10-15
Implement group task sidebar icon colors from course IDs

ptwu:
### Summary 

This pull request implements group task sidebar icon colors as in the Figma design [here](https://www.figma.com/file/MoF3jQ5hQ9dT76iZ57i9VN/Group-Work?node-id=878%3A3099). Previously, groups all had the same color, but this can help users disambiguate different classes with the same subject. This required a refactor of GroupIcon to get the specific tag from the course ID from the Redux store and extract the color of the tag for it. Course IDs for Tag objects are in the form `{some number} {subject} {class number}`, but the `classCode` given to the `GroupIcon` is only `{subject} {class number}`. We split the string to compare only the subject and class number to try and match the tag to a course, and update the group icon background color accordingly.

- [x] implemented colors for group icons based on corresponding class tag colors from the user

### Test Plan 

image

Enable group tasks in the GateKeeper. Then create some groups for your classes, and see that the group icons match the colors of the tags in personal view for the same classes.

### DIFF
--- a/frontend/src/components/SideBar/GroupIcon.module.scss
+++ b/frontend/src/components/SideBar/GroupIcon.module.scss
@@ -1,7 +1,6 @@
 @import '../../colors.scss';
 
 .GroupIcon {
-  background: $aqua;
   line-height: 44px;
   text-align: center;
   font-size: 32px;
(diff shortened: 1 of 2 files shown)


SamChou19815 (approved):
LGTM for this one.

I hope there can be a refactor later that doesn't make the implementation depend on the form of `classId`. Ideally we should treat this as an opaque value and only use it to fetch data. This means that we might have to change what we store in classCode.

PR #593, opened 2020-10-15
Support editing group name

SamChou19815:
### Summary 

This diff added the support for editing group name. The [design](https://www.figma.com/file/MoF3jQ5hQ9dT76iZ57i9VN/Group-Work?node-id=878%3A820) doesn't seem to tell us how it should look, so I decided to go with the prompt modal first. We already have a prompt framework setup, so even if we don't end up using prompt for name edit, we still don't waste too many time on the UI code.

It turns out that the code for ViewSelector also needs some revision. Right now we store the entire group object in the state. This is bad because when the group from firestore changes, we will end up storing an stale value. Thus, I changed it to store selected group id instead, so that when content of the group updates (e.g. name), it will correctly trigger an rerender.

### Test Plan 

edit

### Note

You can easily see that there is a flash in the test plan. This is not caused by this diff but instead by some bad setup that caused react to fail to cache the data between rerenders. An interest problem to solve for next week if anyone wants to take the challenge.

### DIFF
--- a/frontend/src/ViewSwitcher.tsx
+++ b/frontend/src/ViewSwitcher.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from 'react';
-import type { Group, State } from 'common/types/store-types';
+import type { State } from 'common/types/store-types';
 import { useSelector } from 'react-redux';
 import SideBar from './components/SideBar';
 import PersonalView from './PersonalView';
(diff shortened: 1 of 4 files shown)


ptwu (commented):
image

Maybe add a `cursor: pointer` attribute when hovering so users know that it's clickable.

Also, is there some filled pencil SVG we have so that the icon also changes on hover?

SamChou19815:
- Removed space to make hover css work.
- Added `opacity: 0.8` to show some highlighting effect

ptwu (approved):

PR #594, opened 2020-10-15
Group TaskEditor UI

### LINKED ISSUE #580: [group-task] Make TaskEditor to fit the design for group tasks
The task editor for group task should look similar to the design [here](https://www.figma.com/file/MoF3jQ5hQ9dT76iZ57i9VN/Group-Work?node-id=878%3A3099). Unlike the design, I think we should still keep the delete button.

ptwu:
### Summary 

This PR implements the group task editor UI according to the designs [here](https://www.figma.com/file/MoF3jQ5hQ9dT76iZ57i9VN/Group-Work?node-id=878%3A3099). It keeps the delete button for tasks and subtasks. We also make some changes to the default TaskEditor (small font and size changes) to make sure design is consistent across the app. The bell/nudge button still has no functionality.

- [x] implements the new task editor UI design for group tasks
- [x] fixes #580

### Test Plan 

RK8tgGt

Enable group tasks through the `GateKeeper` and create a group. Since we don't have UI functionality for adding group members or assigning tasks to group members yet, create some dummy group tasks in the Firebase cloud console and see that they conform to the design.


frontend/src/components/SideBar/GroupIcon.tsx
```diff
@@ -1,21 +1,41 @@
...
+const GroupIcon = ({ classCode, handleClick, selected, tags }: Props): ReactElement => {
+  const currentClassTag =
+    tags.find(({ classId }) => {
+      const segments = classId !== null ? classId.split(/\s+/) : null;
+      if (segments === null) {
+        return false;
+      }
+      const classIdWithoutNumerical = `${segments[1]} ${segments[2]}`;
+      return classIdWithoutNumerical === classCode;
+    }) ?? NONE_TAG;
```
SamChou19815, line 26: I have seen similar code like this in #590.
Not push blocking for now, but I think it's worth to factor it out to be a custom hook, using the [`useSelector` hook](https://react-redux.js.org/api/hooks#useselector-examples).

SamChou19815 (commented):
Overall looks reasonable. Not approving now to give other people a chance to look at it. @cornell-dti/samwise

frontend/src/components/SideBar/GroupIcon.tsx
SamChou19815: oh just realized that this diff contains the already merged commit from #590. Merging in latest master or rebasing can make this change disappear.

frontend/src/components/Util/TaskEditors/TaskEditor/index.tsx
```diff
@@ -52,6 +52,8 @@ type OwnProps = DefaultProps & {
   readonly subTasks: readonly SubTask[];
   readonly actions: Actions; // The actions to perform under different events
   readonly calendarPosition: CalendarPosition;
+  readonly isGroupTask?: boolean;
```
pihucode, line 55: I think you can get this indirectly by checking if `memberName` is defined.

frontend/src/components/Util/TaskEditors/InlineTaskEditor.tsx
```diff
@@ -9,6 +9,8 @@ type Props = {
   readonly filtered: Task;
   readonly calendarPosition: CalendarPosition;
   readonly className?: string; // additional class names applied to the editor.
+  readonly isGroupTask?: boolean;
```
pihucode, line 12: Here, too. (See above comment)

pihucode (approved):
The UI looks good on my end! One minor issue is that the strikethrough is not centered, but I don't think that should block merging.

strikethrough

ptwu:
> The UI looks good on my end! One minor issue is that the strikethrough is not centered, but I don't think that should block merging.
> 
> strikethrough

This might be more because of the font change, but I'll look into it.

PR #595, opened 2020-10-15
Update course-info script for FA20 semifinal data

### LINKED ISSUE #510: Make course-info script fetch FA20 final info when Cornell releases it

ptwu:
### Summary 

This PR implements the final data parser, adjusted for the FA20 semifinal schedule released today [here](https://registrar.cornell.edu/calendars-exams/semifinal-exam-schedule). It has been adapted so that it fits this semester's current layout.

- [x] gets one step closer to getting rid of #510, but COVID sucks

### Test Plan 

Run the bash script in `/course-info` by running `./run` in the directory. See that no errors occur. You can cross-check the generated .json file to see if the examTimes field for a certain class object is defined. For example, CS 3110 has a semifinal on Nov 18 at 7:30PM, and we can cross-check the Unix time to see it is the same date.

image

### DIFF
--- a/course-info/src/fetch-exam.ts
+++ b/course-info/src/fetch-exam.ts
@@ -3,7 +3,7 @@ import { JSDOM } from 'jsdom';
 import { ExamInfo } from './types';
 
 const prelimUrl = 'https://registrar.cornell.edu/exams/fall-prelim-exam-schedule';
-const finalUrl = 'https://registrar.cornell.edu/exams/fall-final-exam-schedule';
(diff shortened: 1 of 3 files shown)


SamChou19815:
I think final exam is still a thing, so we should probably attach the times to a new `semiFinal` field.

SamChou19815 (approved):
Looks good. Then you only need to update the frontend to use the semifinal field. You can do it in this PR or open up another one.

PR #596, opened 2020-10-16
update firestore group type and update listeners

jason-tung:
### Summary 

Correct the type of FirestoreGroup from Date to firestore.Timestamp.

### Test Plan 

Play test on staging but should not break anything


SamChou19815 (approved):
lgtm

PR #598, opened 2020-10-16
Add semifinal to exam + time type in frontend 

ptwu:
### Summary 

This pull request adds semifinals as a type of exam paired with its time in the frontend so we can display it as imported exams.

### Test Plan 

:eyes:


SamChou19815 (approved):
powertothetypes

Issue #577, opened 2020-10-10
[group-task] Implement progress bar


SamChou19815:
- [x] Compute the progress for the progress
- [x] Render the beautiful progress bar

SamChou19815:
Close since it's good enough for release. Animations can be added later.

PR #601, opened 2020-10-17
Group progress bar

jason-tung:
### Summary 
Not completely done implementing all the design on Figma, but the rough functionality is there.

What I do have: The progress bar has the "x days til deadline" completely working, the flag and deadline on the right hand side, and the bear image when there are no tasks, and properly tracks the number of tasks and completed tasks and creates the appropriate bubbles.

What I still need to do: Animate completing group tasks so the bubble does not just suddenly turn green without a transition, implement logic for 20+ tasks.

grouptaskprogress

works towards #577 

### Test Plan 

Playtest on staging :)


frontend/src/components/GroupView/index.tsx
```diff
@@ -56,13 +56,15 @@ const useGroupMemberProfiles = (
...
   const [groupTaskArray, setGroupTaskArray] = useState<Task[]>([]);
+  const [showBar, setShowBar] = useState<boolean>(false);
   const groupID = group.id;
   useEffect(() => {
     database
       .tasksCollection()
       .where('type', '==', 'GROUP')
       .where('group', '==', groupID)
       .onSnapshot((snapshot) => {
+        setShowBar(snapshot.docs.length > 0);
```
SamChou19815, line 67: Why can't we just do `const showBar = groupTaskArray.length > 0`?

frontend/src/components/GroupView/RightView/GroupTaskProgress.tsx
```diff
@@ -0,0 +1,82 @@
...
+    'Sep',
+    'Oct',
+    'Nov',
+    'Dec',
+  ];
+  const oneDay: number = 1000 * 60 * 60 * 24;
+  // Always round down the number of days left. i.e. if the number of hours
+  // left is <24 just say there are 0 days left
+  const daysLeft: number = Math.floor(
+    Math.max((deadline.getTime() - new Date().getTime()) / oneDay, 0)
```
SamChou19815, line 52: This is potentially dangerous due to daylight saving.

ptwu (approved):

PR #605, opened 2020-10-17
Fix overdue alert style so it appears on task card

### LINKED ISSUE #604: [bug][task] Overdue task `!` badge should be top-right of the card, not the entire window
**Describe the bug**

Title

**To Reproduce**

Have an overdue task

**Screenshots**

Screen Shot 2020-10-17 at 17 02 02

**Environment (optional)**
 - Browser [e.g. chrome, safari] Chrome
 - OS: [e.g. macOS] macOS

**Additional context (optional)**
:(

ptwu:
### Summary 

This PR fixes the overdue task `!` badge to place it in the top right again. This bug is currently in prod, so we should maybe consider releasing soon.

- [x] fixes #604

### Test Plan 

Make a task and make it overdue. See that the warning symbol is now at the top right of the card. Check both group and personal view and see it works.
Before
image

After
image
image

### DIFF
--- a/frontend/src/components/UI/OverdueAlert.module.scss
+++ b/frontend/src/components/UI/OverdueAlert.module.scss
@@ -1,7 +1,7 @@
 .OverdueAlertInTaskCard {
-  position: absolute;
+  position: relative;
   top: -5px;
-  right: -5px;
+  left: calc(100% - 20px);
   width: 2em;
   height: 2em;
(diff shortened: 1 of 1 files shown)


SamChou19815 (approved):

ptwu:
will try to fix this in a more principled way this week

Issue #581, opened 2020-10-10
Migrate all string owner to string array owner


SamChou19815:
We need string array owner for all tasks to ensure that we can have a single, unified query that can fetch all tasks that can be displayed in personal view.

- [x] Create a script that can convert all existing data to string array.
- [x] Fix relevant code in the `frontend` package
- [x] Fix relevant code in the `functions` package

PR #607, opened 2020-10-22
Firestore Task owner schema function (first step)

ptwu:
### Summary 

This pull request implements the first step of fixing #581, or migrating the current Task `owner` schema of a string to an array. The current plan is to see the dev server run this function, and then everything should break, but I'll try to introduce a change quickly for refactoring the frontend to respond to the `owner` change as well as some small changes in the functions

Some implementation details: I based the script off of the remove old tasks function, but refactored the `partition` function which circumvents the batch write limit of 500 to accomodate different types using generics. This is now located in a `util.ts` file. We might want to include some more functions in here later for other functions, so I bypassed some linter rules for it.

Mainly looking for reviewers to see if the code makes sense / sanity check and should update the field accordingly without causing further issues. 

### Test Plan 

Run the script after the PR is merged in. 

### Breaking Changes  

**yes**

This PR introduces a fundamental change to the Task schema. As such, it's expected that all functionality related to Tasks should break. However, this PR depends on a respective frontend change being created first, so it's expected that there will be no difference to a user on the frontend.


common/types/firestore-types.ts
```diff
@@ -6,6 +6,11 @@ export type FirestoreCommon = {
   readonly order: number;
 };
 
+export type FirestoreCommonNew = {
```
ptwu, line 9: This type is currently unused. The goal is to eventually replace the current type with this after the cron job runs.

common/types/firestore-types.ts
```diff
@@ -6,6 +6,11 @@ export type FirestoreCommon = {
...
+  readonly owner: string[];
```
SamChou19815, line 10: use `readonly string[]` for immutability

functions/src/index.ts
```diff
@@ -8,3 +9,5 @@ export const iCalFunction = functions.pubsub.schedule('0 0 * * *').onRun(getICal
 export const FocusTasksDueToday = functions.pubsub.schedule('0 0 * * *').onRun(focusTasksDueToday);
 
 export const RemoveOldTasks = functions.pubsub.schedule('0 0 * * *').onRun(removeOldTasks);
+
+export const MigrateTaskOwners = functions.pubsub.schedule('0 0 * * *').onRun(migrateTaskOwners);
```
SamChou19815, line 13: I don't think it should be a cron job. I think it's just something we should run once when no one is using it.

SamChou19815:
BTW, this will probably leave unmerged until a frontend code change is done.

common/types/firestore-types.ts
ptwu: > BTW, this will probably leave unmerged until a frontend code change is done in another PR.

sounds good. I thought it might be easier to fix things once the schema was already changed, but this is definitely the safer option.

ptwu:
> BTW, this will probably leave unmerged until a frontend code change is done in another PR.

sounds good. I thought it might be easier to fix things once the schema was already changed, but this is definitely the safer option.

functions/src/index.ts
ptwu: oops, kind of messed up there. making it a cron job would actually cause errors and make the owner fields multidimensional arrays because the script assumes by default that owners are only strings 😕

SamChou19815 (approved):
look good, maybe you can make `functions/src/migrate-task-owner.ts` directly runnable and run it by `ts-node`

PR #608, opened 2020-10-22
Fix frontend task owner code to accomodate new owner schema

ptwu:
### Summary 

This pull request fixes relevant code in the frontend and in functions that considers task owners to be strings rather than string arrays. We created the script in #607 to run on all tasks in the db and change the owners to singleton string arrays instead of strings, and this corresponding PR allows for the data from the backend to be displayed correctly.

- [x] second part of fixing #581 

### Test Plan 
Review this PR carefully and run the script from #607. If the changes happen as expected in the Firestore, we can merge this in after. See that functionality is preserved as it was for personal view and everything.

### Notes
The default value of `owner` for a task used to be the empty string `''`, but is now `['']`.

### Breaking Changes  

This is a breaking change as it changes the default value of owners in tasks, and how owners are handled in the frontend.

### DIFF
--- a/common/types/firestore-types.ts
+++ b/common/types/firestore-types.ts
@@ -2,7 +2,7 @@ import { firestore } from 'firebase/app';
 import { RepeatingPattern } from './store-types';
 
 export type FirestoreCommon = {
-  readonly owner: string;
+  readonly owner: string[];
   readonly order: number;
 };
 
--- a/common/types/firestore-types.ts
+++ b/common/types/firestore-types.ts
@@ -2,7 +2,7 @@ import { firestore } from 'firebase/app';
 import { RepeatingPattern } from './store-types';
 
 export type FirestoreCommon = {
-  readonly owner: string[];
+  readonly owner: readonly string[];
   readonly order: number;
 };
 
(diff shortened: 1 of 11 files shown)


common/types/firestore-types.ts
```diff
@@ -2,7 +2,7 @@ import { firestore } from 'firebase/app';
 import { RepeatingPattern } from './store-types';
 
 export type FirestoreCommon = {
-  readonly owner: string;
+  readonly owner: string[];
```
SamChou19815, line 5: `readonly string[]`

SamChou19815 (commented):
You also need to change this function code:

https://github.com/cornell-dti/samwise/blob/289219f8e3f82e7c9037ecc6a0aa0dd2d590c396/functions/src/iCalFunctions.ts#L38-L49

otherwise looks good

ptwu:
> You also need to change this function code:
> 
> https://github.com/cornell-dti/samwise/blob/289219f8e3f82e7c9037ecc6a0aa0dd2d590c396/functions/src/iCalFunctions.ts#L38-L49
> 
> otherwise looks good

It looks changed for me :/ That code block is from `master`.

common/types/firestore-types.ts
ptwu: oops, fixed!

SamChou19815 (approved):
oops missed that one. LGTM.

Nits: change all `string[]` in your changes to `readonly string[]`

PR #497, opened 2020-05-08
Backend functions for group tasks

pihucode:
### Summary

Implements backend functions for managing group tasks and assigning members to tasks. At the moment group tasks can be assigned to only one member. Group tasks are displayed in personal view, but can be tackled in a future PR. When initially creating a group task, members cannot be assigned due to a firebase permission error, however, members can be edited once the group task is created.

- [x] create group task
- [x] edit group task
- [x] delete group task
- [x] edit member for a group task

### Test Plan 

Tested manually and looked at changes in firebase console.
To be tested more once frontend features are working.

### Notes

Added an `owner` field to `MainTask` and `TaskMetadata`.

PR #611, opened 2020-10-23
Fix frontend to work as expected after Task owner migration

### LINKED ISSUE #581: Migrate all string owner to string array owner

ptwu:
### Summary 

Merging in #607 and #608 and applying the Firebase function from #607 results in some issues. namely, the change to the `FirestoreCommon` type, which assumes that every firestore document now has an owner field of type `string[]`. However, this only applies to tasks - other entities will still have singular owners. As a result, subtasks and tags were not being shown.

This PR also introduces new security rules to fit the new `owner` definition for tasks, and fixes the `mergeWithOwner` function for subtasks, since subtasks still have singular owners.

- [x] fixes #581

### Test Plan 

See that frontend on staging should display everything as normally intended: subtasks for tasks with correct tags, group icons have the correct background colors:

image

### DIFF
--- a/common/types/firestore-types.ts
+++ b/common/types/firestore-types.ts
@@ -2,7 +2,7 @@ import { firestore } from 'firebase/app';
 import { RepeatingPattern } from './store-types';
 
 export type FirestoreCommon = {
-  readonly owner: readonly string[];
+  readonly owner: string | readonly string[];
   readonly order: number;
 };
 
--- a/common/firebase/common-actions.ts
+++ b/common/firebase/common-actions.ts
@@ -19,7 +19,10 @@ export default class Actions {
     source: T
   ): Promise<T & FirestoreCommon> => {
     const order = await this.orderManager.allocateNewOrder(orderFor);
-    return { ...source, owner: [this.getUserEmail()], order };
+    if (orderFor === 'tasks') {
+      return { ...source, owner: [this.getUserEmail()], order };
+    }
+    return { ...source, owner: this.getUserEmail(), order };
   };
 
   /*
--- a/common/util/general-util.test.ts
+++ b/common/util/general-util.test.ts
@@ -31,4 +31,6 @@ it('shallowArrayEqualWorks', () => {
   ).toBeTruthy();
   expect(shallowArrayEqual([{ foo: 'bar', bar: 'foo' }], [])).toBeFalsy();
   expect(shallowArrayEqual([{ foo: 'bar', bar: 'foo' }], [{ foo: 'bar' }])).toBeFalsy();
+  expect(shallowArrayEqual([''], [''])).toBeTruthy();
+  expect(shallowArrayEqual([3, 4, 5, 6], [3, 4, 7, 3])).toBeFalsy();
 });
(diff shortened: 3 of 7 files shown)


SamChou19815:
@ptwu I think the creating task doesn't work is due to this bad default value:

https://github.com/cornell-dti/samwise/blob/5408801f9a7133b27fa48477978a8ea0e77fd767/frontend/src/components/TaskCreator/index.tsx#L58

and there is an editOwner function that is never called:

https://github.com/cornell-dti/samwise/blob/5408801f9a7133b27fa48477978a8ea0e77fd767/frontend/src/components/TaskCreator/index.tsx#L200-L206

ptwu:
> @ptwu I think the creating task doesn't work is due to this bad default value:
> 
> https://github.com/cornell-dti/samwise/blob/5408801f9a7133b27fa48477978a8ea0e77fd767/frontend/src/components/TaskCreator/index.tsx#L58
> 
> and there is an editOwner function that is never called:
> 
> https://github.com/cornell-dti/samwise/blob/5408801f9a7133b27fa48477978a8ea0e77fd767/frontend/src/components/TaskCreator/index.tsx#L200-L206

I removed the `editOwner` function because it was unused, and it assumes that the `owner` field is a singular value. Checking out the git blame shows that it was introduced by Pi here in #497. 

The old default value was the empty string. I don't think this is the problem since the default value is handled in `actions.ts` here
https://github.com/cornell-dti/samwise/blob/9b0ede85825aaf0c5346b3c3239ae1c99d6e5d74/frontend/src/firebase/actions.ts#L127

My guess is that there might be an error in the security rules for creating a task.

SamChou19815:
@ptwu here is the problem:

https://github.com/cornell-dti/samwise/blob/9b0ede85825aaf0c5346b3c3239ae1c99d6e5d74/frontend/src/firebase/actions.ts#L127

Screen Shot 2020-10-23 at 12 23 27

ptwu:
> @ptwu here is the problem:
> 
> https://github.com/cornell-dti/samwise/blob/9b0ede85825aaf0c5346b3c3239ae1c99d6e5d74/frontend/src/firebase/actions.ts#L127
> 
> Screen Shot 2020-10-23 at 12 23 27

image

SamChou19815:
Loop in @pihucode as well. I think it's a bad idea to use `''` as the special value to decide whether we should use your own email as default owner. It is not obvious from the type signature. A better alternative is to use `readonly string[] | null` or `readonly string[] | undefined` where the `null` case means that we should use default owner.

ptwu:
> Loop in @pihucode as well. I think it's a bad idea to use `''` as the special value to decide whether we should use your own email as default owner. It is not obvious from the type signature. A better alternative is to use `readonly string[] | null` or `readonly string[] | undefined` where the `null` case means that we should use default owner.

We can introduce this in an upcoming PR, since the goal here is to just get things working again 😢.

common/types/firestore-types.ts
```diff
@@ -2,7 +2,7 @@ import { firestore } from 'firebase/app';
 import { RepeatingPattern } from './store-types';
 
 export type FirestoreCommon = {
-  readonly owner: readonly string[];
+  readonly owner: string | readonly string[];
```
SamChou19815, line 5: I can accept this for now to unbreak dev. But we should really move `owner` out of `FirestoreCommon`. Otherwise I will be able to put a string array into a `FirestoreTag`.

SamChou19815 (approved):
only one non push blocking nit, otherwise LGTM

PR #613, opened 2020-10-23
Setup FirestoreDataConverter for samwise-settings collection

SamChou19815:
### Summary 

Setup infra for using [`FirestoreDataConverter`](https://firebase.google.com/docs/reference/js/firebase.firestore.FirestoreDataConverter) to strictify the type enforcement when we are moving between firestore data and app data. Previous, this layer is horribly `any`-typed. This diff is the first step to change that.

To avoid large disruptions, I started with an easy target: `samwise-settings`. The schema for this collection is very simple, and it's not used very often.

### Test Plan 

Go to staging to check that it still remembers that you finished onboarding tutorial, which shows that settings are not broken.

### DIFF
--- a/common/firebase/database.ts
+++ b/common/firebase/database.ts
@@ -1,5 +1,7 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
+import { Settings } from '../types/store-types';
+
 type DocumentData = { [field: string]: any };
 type UpdateData = { [fieldPath: string]: any };
 type WhereFilterOp =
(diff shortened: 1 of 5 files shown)


common/firebase/database.ts
```diff
@@ -111,13 +120,37 @@ export interface CommonFirestore {
...
+      typeof canvasCalendar !== 'undefined' &&
+      canvasCalendar !== null &&
+      typeof canvasCalendar !== 'string'
+    ) {
+      throw new Error();
+    }
+    if (typeof completedOnboarding !== 'boolean') throw new Error();
+    if (theme !== 'light' && theme !== 'dark') throw new Error();
+    return { canvasCalendar, completedOnboarding, theme };
+  },
```
SamChou19815, line 143: we do some runtime type validation here to ensure that we never give the rest of the system incorrectly formatted object.

ptwu (approved):

PR #614, opened 2020-10-24
Unbreak course-info semifinal exam script for CS 2800

ptwu:
### Summary 

This PR unbreaks the course exam script which was erroring on one line: CS 2800. This is because [the registrar](https://registrar.cornell.edu/calendars-exams/semifinal-exam-schedule) seems to have manually inputted this line without a hyphen in the date of the exam, which is different from the original format of each line. We hardcode the exam data for CS 2800 to avoid an error on that line.

### Test Plan 

image

### Notes 

lol


SamChou19815 (approved):
we will blame the registrar for this hardcoding

PR #616, opened 2020-10-24
Unbreak group task creation

SamChou19815:
### Summary 

We currently have a bug that let's you create a group task, but the group task just won't appear in your group view.
It's because we give an incorrect value for the `group` field in metatada. `group` should store group id instead of group name.

This diff fixes this issue.

### Test Plan 

Add a group task in group view and it should appear in your group view.


meganyin13 (approved):
LGTM 🚀

PR #619, opened 2020-10-25
Replace hacky gh-pages deploy with firebase deploy preview

SamChou19815:
### Summary 

One year after I introduced the hacky gh-pages based deploy solution to create a staging URL for each PR, firebase finally releases the official solution. This PR tries to adopt that.

### Test Plan 

CI


meganyin13 (approved):
🔥🔥🔥

SamChou19815:
The gh-pages will still be kept around but branch protected to avoid further modification, so that @SamChou19815 and @meganyin13's github contribution graph won't be broken 😃

Issue #339, opened 2019-10-27
Tag Field in Task Editor Doesn't Look Clickable


mt-xing:
**Describe the bug**
Users don't realize that the tag name inside the task editor component can be clicked, leading to confusion about how to change the tag of a preexisting task.

**Screenshots**
The MATH 3320 field is clickable:
![image](https://user-images.githubusercontent.com/6147405/67640241-b92a8100-f8cf-11e9-99e8-1988cd9e255f.png)
But you have to hover over it to notice:
![image](https://user-images.githubusercontent.com/6147405/67640238-a617b100-f8cf-11e9-8321-067d5db30d3e.png)

SamChou19815:
We lost all the designers and I'm not sure how we plan to do with this. Closing it unless we suddenly got some design.

PR #621, opened 2020-10-28
Fix Task Tag Text Size

meganyin13:
### Summary
We have a regression in some release where the task tag text just got bigger. Looks to be due to [this line](https://github.com/cornell-dti/samwise/blob/68ac210a35dff04d1f89473252ff147320b19a42/frontend/src/components/Util/TaskEditors/TaskEditor/index.module.scss#L113) in TaskEditorTag introduced by #594. Not sure why it was added but removing it doesn't break anything in group/personal view and reverts the task tag to the intended size 🤷🏻‍♀️

### Test Plan

Before:
prod

After:
dev

### DIFF
--- a/frontend/src/components/Util/TaskEditors/TaskEditor/index.module.scss
+++ b/frontend/src/components/Util/TaskEditors/TaskEditor/index.module.scss
@@ -110,7 +110,6 @@ img.TaskEditorIcon:focus,
 
 .TaskEditorTag {
   margin-left: 0.5em;
-  font: 1.3em 'Open Sans', sans-serif;
   padding: 5px;
   border-radius: 2px;
   color: white;


frontend/src/components/Util/TaskEditors/TaskEditor/index.module.scss
```diff
@@ -110,7 +110,6 @@ img.TaskEditorIcon:focus,
 
 .TaskEditorTag {
   margin-left: 0.5em;
-  font: 1.3em 'Open Sans', sans-serif;
```
ptwu, line 113: I added this as part of the [Group Task redesign](https://www.figma.com/file/MoF3jQ5hQ9dT76iZ57i9VN/Group-Work?node-id=878%3A3099), where the tasks have Open Sans font. The main regression we noticed is that the floating task editor has bigger font for some reason even compared to tasks in focus view.

ptwu: maybe removing the font size attribute will do the trick?

meganyin13: wait but we don't have tags in group tasks?

ptwu: [It's referring to this tag field](https://github.com/cornell-dti/samwise/issues/339)

ptwu (approved):
LGTM, it solves the issue with floating task editor size and we can just use the old task field design.

PR #623, opened 2020-10-31
Firestore Task <- Subtask merge function (step 1) + fix DB credential

ptwu:
### Summary 

This PR is the first step towards implementing the subtask and task merge. We first grab all the subtasks in the `samwise-subtasks` collection and consolidate them in an immutable.js Map, indexed by their ID. Then, we iterate through all tasks in the `samwise-tasks` collection and build an array of temporary objects representing subtasks and their corresponding Task ID. Lastly, we do a batch update of the tasks with this subtask data, replacing children as an array of strings with an array of schema of the old subtask. (array of `{ name: string, complete: boolean, inFocus: boolean}`)

### Test Plan 

**DEPENDENT ON MERGING IN A FRONTEND CHANGE FIRST**

Run the function using `ts-node`. Optionally, run a verification function that checks that all children match the subtask schema, and are an array of them. Then, if the function runs successfully, we can delete the `samwise-subtasks` collection.

### Notes

The code currently doesn't compile since we haven't implemented the frontend changes to respond to the new subtask in task schema. This will be an involved process, and after a frontend change is merged in, we'll merge `master` in this PR and run the script soon.

### Breaking Changes  

This PR introduces a fundamental change to the Task schema. As such, it's expected that all functionality related to Tasks should break. However, this PR depends on a respective frontend change being created first, so it's expected that there will be no difference to a user on the frontend.


SamChou19815 (commented):
script looks good, will approve once the frontend changes are done and the type error is fixed

functions/src/merge-task-subtask.ts
```diff
@@ -0,0 +1,55 @@
...
+const mergeSubtaskTask = async (): Promise<void> => {
+  const tasksSnapshot = await database.tasksCollection().get();
+  const subTasksSnapshot = await database.subTasksCollection().get();
+  // Initialize map from subtask ID to FirestoreCommonTask representing its data
+  let subTaskMap: Map<string, FirestoreSubTask> = Map();
+
+  // Populate map for constant time retrieval of subtasks by ID
+  subTasksSnapshot.forEach((document) => {
+    const data = document.data();
+    const subTask = data as FirestoreSubTask;
```
SamChou19815, line 17: In #635, you removed id from subtask. This cast will hide the problem that you forget to remove id from subtask.

SamChou19815 (changes requested):
request changes so you don't forget to fix it.

functions/src/merge-task-subtask.ts
```diff
@@ -0,0 +1,83 @@
...
+                [...subTaskOrderSet].reduce((acc, curr) => Math.max(acc, curr), 0) + 1;
+              subTaskOrderSet.add(uniqueOrder);
+              subtasks.push({ complete, inFocus, name, order: uniqueOrder });
+            } else {
+              subTaskOrderSet.add(order);
+              subtasks.push({ complete, inFocus, name, order });
+            }
+          }
+        });
+      }
```
SamChou19815, line 64: Not sure why we need this order set, but it looks harmless here. I guess it's because in the past we somehow messed up the order of subtasks?

SamChou19815 (approved):
one little question, otherwise looks good

functions/src/merge-task-subtask.ts
ptwu: Yeah, there are some tasks with subtasks that have identical orders. Since it's necessary that orders are unique within a task now, we have to ensure this so users don't have corrupted data.

PR #625, opened 2020-10-31
Unbreak getDateWithDateString to fix #624

### LINKED ISSUE #624: Subtasks on Repeating Tasks in Future View are Broken
**Describe the bug**
Unable to check off subtasks on repeated tasks in future view.
In fact, the whole subtask system seems to be broken on repeated tasks - trying to focus a subtask results in it showing up in focus view but not updating the pin icon in future view, allowing me to click the pin icon a second time, resulting in a second copy of the same subtask showing up in focus.

**To Reproduce**
Have a repeated task in future view. Do not focus the task at all. Try to check off a subtask or pin the subtask to focus. Observe broken.

**Screenshots**
ezgif com-gif-maker

After clicking the pin icon many times:
image
(note the repeats inside focus)

**Environment (optional)**
 - Browser [e.g. chrome, safari] Chrome
 - OS: [e.g. macOS] Windows 10
This is running in production.

SamChou19815:
### Summary 

It turns out that `getDateWithDateString` has never been correct. It's always slightly wrong in some weird ways. This is what happens:

https://github.com/cornell-dti/samwise/blob/72aeb1ddc75401244758d090f9c0c1f8d400807c/common/util/datetime-util.ts#L86-L97

Today is Oct 31. Assume we want to set date for a time string `'Tue Nov 03 2020'`.
When we call `date.setMonth(10)`, the following things might be happening under the hood:

1. JS figure out that we need to move one month forward.
2. However, after we move one month forward, we landed in Nov. 31, which doesn't make sense.
3. To fix this, JS turns Nov. 31 into Dec 1.

Then when we set the date, we got this epic OFF-BY-ONE-MONTH error 😱 , which is the underlying cause of #624.

This diff tries to `setDate(1)` first to avoid this kind of tricky conversion. In the long term, we should really consider using a datetime library.

### Test Plan 

`yarn test`. The added test will fail before this change.

### DIFF
--- a/common/util/datetime-util.test.ts
+++ b/common/util/datetime-util.test.ts
@@ -1,4 +1,10 @@
-import { day2String, isToday, getTodayAtZeroAM, getDateAfterXWeeks } from './datetime-util';
+import {
+  day2String,
+  isToday,
+  getTodayAtZeroAM,
+  getDateWithDateString,
+  getDateAfterXWeeks,
+} from './datetime-util';
 
 it('day2String works', () => {
   expect(day2String(0)).toBe('SUN');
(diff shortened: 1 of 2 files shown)


ptwu (approved):
LGTM. Hopefully no half-day months are introduced in the future.

PR #626, opened 2020-10-31
Fix scrollbars and CSS in GroupView

pihucode:
### Summary

This pull request fixes scrollbars and styling in `GroupView`. Tasks appearing in `TaskQueue` are scroll-able. The container for search results is given a max height to avoid results going offscreen and is made scroll-able. The scrollbar in `GroupTaskContainer` has been fixed so now no group tasks are cut off from the right side of the window. `RightView` should cover the remaining window space so its height is set to `calc(100vh - TaskCreator height - GroupTaskProgress height)` and its width is set to `calc(100vw - SideBar width - MiddleBar width)`.

- [x] Added scrollbars to `TaskQueue `and search results in assigning members
- [x] Fixed scrollbar in `GroupTaskContainer`

### Test Plan

Test scrollbars on UI
scrollbars

Placed dummy values in search results to test scrollbar
scroll search res

### Notes
Dummy values used in testing `TaskQueue` and search results are not pushed.


frontend/src/components/GroupView/RightView/index.module.scss
```diff
@@ -5,7 +5,9 @@
 .RightView {
   display: flex;
   flex-direction: column;
-  height: 480px;
+  height: calc(100vh - 140px);
+  width: calc(100vw - 80px - 280px);
```
SamChou19815, line 9: We should move the width and height of group view stuff into a constant file, and directly reference that constant.

frontend/src/components/GroupView/MiddleBar/GroupViewMiddleBarTaskQueue.module.scss
```diff
@@ -30,3 +30,8 @@
   cursor: pointer;
   opacity: 0.7;
 }
+
+.TaskList {
+  max-height: 85%;
```
SamChou19815, line 35: why this weird percentage?

frontend/src/components/GroupView/RightView/index.module.scss
pihucode: Fixed this! Moved widths and heights into `groupview-sizes.scss`

frontend/src/components/GroupView/MiddleBar/GroupViewMiddleBarTaskQueue.module.scss
pihucode: This was so that the rendered task list does not overflow into the `People` section. I just changed its height values in a recent commit.

ptwu (commented):
image

weird regression i'm experiencing on Windows 10 (chrome).

ptwu (approved):
LGTM

Issue #618, opened 2020-10-24
[code-quality] TaskCreator css code is terrible


SamChou19815:
Looks like we added a bunch of `positive: absolute` for elements that don't really need them. As a result, the code is very fragile:

<img width="1792" alt="Screen Shot 2020-10-24 at 17 46 31" src="https://user-images.githubusercontent.com/4290500/97094219-df891680-1620-11eb-9129-ac1f42f2dc51.png">

Looks like all we want to do is to workaround the issue that the `input` is the first element within form.

https://github.com/cornell-dti/samwise/blob/c24cc059431f61a12a22dbf870e4785905840bea/frontend/src/components/TaskCreator/index.tsx#L490-L507

Then we started to use `position: absolute` so that we can actually place something before the `input` 😢 🤦

SamChou19815:
#617 starts my attempt to fix these css issues. There will be a lot more coming.

PR #628, opened 2020-10-31
Unify this.state.opened and this.props.taskCreatorOpened into isOpen

SamChou19815:
### Summary 

Add this helper function to be consistent when we are checking whether the task editor should be open.

Another preparation diff to fix the problem in #618.

### Test Plan 

👀

### DIFF
--- a/frontend/src/components/TaskCreator/index.tsx
+++ b/frontend/src/components/TaskCreator/index.tsx
@@ -85,6 +85,11 @@ export class TaskCreator extends React.PureComponent<Props, State> {
     };
   }
 
+  private get isOpen(): boolean {
+    // eslint-disable-next-line react/destructuring-assignment
+    return this.state.opened || this.props.taskCreatorOpened || false;
+  }
+
   /*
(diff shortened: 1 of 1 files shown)


frontend/src/components/TaskCreator/index.tsx
```diff
@@ -85,6 +85,11 @@ export class TaskCreator extends React.PureComponent<Props, State> {
     };
   }
 
+  private get isOpen(): boolean {
+    // eslint-disable-next-line react/destructuring-assignment
+    return this.state.opened || this.props.taskCreatorOpened || false;
```
ptwu, line 90: Isn't this equivalent to `this.state.opened || this.props.taskCreatorOpened`?

ptwu (approved):
LGTM with one nit

frontend/src/components/TaskCreator/index.tsx
SamChou19815: this.props.taskCreatorOpened can be undefined, so we need the final or case at the end.

PR #630, opened 2020-11-01
Do not share code between isOpen and !isOpen for TaskCreator

SamChou19815:
### Summary 

Trying to share code between the open state and close state of the TaskCreator is the root source of evil that caused the unnecessary `position: absolute` hack mentioned in #618. In the end, we did manage to share a few lines of code, but at the cost of completely unnecessary `position: absolute` hack.

This diff starts to reverse that decision. In particular, it deals with the `!isOpen` case in one place, and then make the rest of the code only worry about the `isOpen` case. It will make the flow easier to understand, and help us untangle the unnecessary dependency between these two, so that we can eventually remove the `position: absolute` hack without worrying about the `!isOpen` case.

### Test Plan 

Closed state still works:
Screen Shot 2020-10-31 at 20 10 36

Open state still works:
Screen Shot 2020-10-31 at 20 09 11

### DIFF
--- a/frontend/src/components/TaskCreator/index.tsx
+++ b/frontend/src/components/TaskCreator/index.tsx
@@ -75,19 +75,14 @@ export class TaskCreator extends React.PureComponent<Props, State> {
 
   private addTask: HTMLInputElement | null | undefined;
 
-  private darkModeStyle: CSSProperties;
-
-  constructor(props: Props) {
-    super(props);
-    this.darkModeStyle = {
-      background: 'black',
-      color: 'white',
-    };
-  }
-
   private get isOpen(): boolean {
     // eslint-disable-next-line react/destructuring-assignment
     return this.state.opened || this.props.taskCreatorOpened || false;
+  }
+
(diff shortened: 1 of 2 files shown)


frontend/src/components/TaskCreator/index.tsx
```diff
@@ -75,21 +75,16 @@ export class TaskCreator extends React.PureComponent<Props, State> {
...
-    super(props);
-    this.darkModeStyle = {
-      background: 'black',
-      color: 'white',
-    };
-  }
-
   private get isOpen(): boolean {
     // eslint-disable-next-line react/destructuring-assignment
     return this.state.opened || this.props.taskCreatorOpened || false;
```
pihucode, line 80: Why is `false` needed here? I don't think `false` affects what value is returned.

SamChou19815: to make sure it type checks

pihucode (approved):

PR #634, opened 2020-11-02
Remove useless onKeyPress handler in GroupTaskRow

SamChou19815:
### Summary 

This handler is introduced in https://github.com/cornell-dti/samwise/commit/28265c512636eef6c95d838441d91d83a6011b66, when the code is still dummy and uses `div` instead of `button`. Now we switched to button so a11y should automatically work without this extra handler. This diff removes it, with a good side effect to removing another linter warning.

### Test Plan 

Try to press that button with tabbing and enter key.


ptwu (approved):
lgtm

PR #635, opened 2020-11-07
RFC: Merge Task <- SubTask Frontend Refactor

ptwu:
### Summary 

This pull request is the final step in the effort to merge `SubTask`s with `Task`s in the Firestore and frontend. It continues on the efforts of #423 with the rationale that the performance/realtime update benefit of querying separate collections for subtasks related to a certain task doesn't overcome the easiness of reasoning about `SubTask` data directly nested within each `Task`. Linear-time operations on the number of subtasks within each task appears to be sufficient, as we don't expect subtask lists to be very long. As such, we choose to store `SubTask`s as objects inside of the `children` field of each `Task`, and a Firebase function has been created for the migration in #623.

The bulk of the migration involves rolling back some granular subtask operations based off of IDs. We update `SubTask`s through their parent tasks using some higher-order functions on arrays instead of getting subtask IDs directly. 

Unit tests for relevant functions on subtasks have been updated accordingly. 

### Test Plan 

`yarn test` with updated tests on `reducers.ts`, `task-util.ts`, and `actions.ts` all pass after modification. Everything also type-checks.

 We need to test on the dev server (not prod), and since this is a significant refactor, chances are things are going to break. After merging this PR in, we should run merge `master` into #623's branch and run the script to see if things work. Extensively test the frontend and side-effects before merging into release.

### Notes 

More details on the implementation strategy are in this design doc that I started writing before starting this PR:

https://drive.google.com/file/d/1B9MbMZHdWbHWYS8_RfKeViwGSceNdC9n/view?usp=sharing

### Breaking Changes  

This diff removes the need for the `samwise-subtasks` collection in the Firestore. We should be able to just remove those security rules and the collection if the migration goes smoothly.


common/util/task-util.ts
```diff
@@ -161,3 +161,6 @@ export function dateMatchRepeats(
   }
   return dateMatchRepeatPattern(date, pattern);
 }
+
+export const subTasksEqual = (firstSubTask: SubTask, secondSubTask: SubTask): boolean =>
+  JSON.stringify(firstSubTask) === JSON.stringify(secondSubTask);
```
SamChou19815, line 166: Maybe we can use https://github.com/cornell-dti/samwise/blob/29f3126ad192279dfc18804dfd016d5716907d04/common/util/general-util.ts#L38-L56
to compare?

frontend/src/components/TaskView/FocusView/ClearFocus.tsx
```diff
@@ -11,22 +11,27 @@ type Props = {
...
     } else {
       t.children.forEach((s) => {
         if (s.inFocus && s.complete) {
-          subTaskIds.push(s.id);
+          const completedSubtasks = subTasksWithParentTaskId.get(t.id);
+          if (completedSubtasks === undefined) {
+            subTasksWithParentTaskId.set(t.id, [s]);
+          } else {
+            subTasksWithParentTaskId.setIn(t.id, [s, ...completedSubtasks]);
+          }
```
SamChou19815, line 26: I don't think you are doing anything useful here.

`Map()` creates an immutable map. So you actually need to do something like `subTasksWithParentTaskId = subTasksWithParentTaskId.set(t.id, [s])` below to assemble together a map.

frontend/src/components/Util/TaskEditors/InlineTaskEditor.tsx
```diff
@@ -41,7 +41,7 @@ export default function InlineTaskEditor({
       icalUID={icalUID}
       taskAppearedDate={taskAppearedDate}
       className={className}
-      mainTask={{ ...mainTask, date: metadata.date }}
+      mainTask={{ ...mainTask, date: metadata.date, children }}
       subTasks={children}
```
SamChou19815, line 45: not sure why we need to pass children twice?

SamChou19815 (changes requested):
Just tested. Editing subtask seems to be very broken.

ptwu:
> Just tested. Editing subtask seems to be very broken.

seems to be working for me on staging now, could you try again?

ptwu:
@SamChou19815 this PR and #623 are ready for review (merging #623 after this one should get rid of the type errors in #623)

frontend/src/components/Util/TaskEditors/TaskEditor/index.tsx
```diff
@@ -74,7 +73,6 @@ function TaskEditor({
   icalUID,
   taskAppearedDate,
   mainTask: initMainTask,
-  subTasks: initSubTasks,
```
SamChou19815, line 77: Consider changing the name because it also contains sub tasks now

frontend/src/components/Util/TaskEditors/TaskEditor/task-diff-reducer.ts
```diff
@@ -133,41 +70,29 @@ function reducer(state: State, action: Action): State {
 
 export default function useTaskDiffReducer(
   initMainTask: MainTask,
```
SamChou19815, line 135: Same here, give it a better name

frontend/src/components/Util/TaskEditors/TaskEditor/task-diff-reducer.ts
```diff
@@ -60,8 +40,8 @@ export function diffIsEmpty(diff: Diff): boolean {
  * @param subTasks an array of subtask for initial state.
  * @returns the initial state.
  */
-function initializer([mainTask, subTasks]: [MainTask, readonly SubTask[]]): State {
-  return { mainTask, subTasks, prevFullTask: { mainTask, subTasks }, diff: emptyDiff };
+function initializer([mainTask]: [MainTask]): State {
```
SamChou19815, line 43: This destructuring looks unnecessary

frontend/src/firebase/actions.ts
```diff
@@ -100,22 +96,15 @@ const asyncAddTask = async (
   newTaskId: string,
   owner: readonly string[],
   task: TaskWithoutIdOrderChildren,
-  subTasks: WithoutId<SubTask>[],
+  subTasks: readonly SubTask[],
   batch: WriteBatch
-): Promise<{ readonly firestoreTask: FirestoreTask; readonly createdSubTasks: SubTask[] }> => {
+): Promise<{ readonly firestoreTask: FirestoreTask }> => {
```
SamChou19815, line 101: Return the task directly instead of wrapping it in object

frontend/src/components/Util/TaskEditors/FloatingTaskEditor.tsx
```diff
@@ -122,12 +122,12 @@ export default function FloatingTaskEditor({
             type={metadata.type}
             icalUID={icalUID}
             taskAppearedDate={taskAppearedDate}
-            mainTask={{ ...mainTask, date: metadata.date }}
-            subTasks={children}
+            taskData={{ ...taskData, date: metadata.date, children }}
```
SamChou19815, line 125: in code above, children is destructured out but not used anywhere, so it's better not separate it from main taskData

frontend/src/components/Util/TaskEditors/InlineTaskEditor.tsx
```diff
@@ -41,8 +41,7 @@ export default function InlineTaskEditor({
       icalUID={icalUID}
       taskAppearedDate={taskAppearedDate}
       className={className}
-      mainTask={{ ...mainTask, date: metadata.date }}
-      subTasks={children}
+      taskData={{ ...taskData, date: metadata.date, children }}
```
SamChou19815, line 44: same here

frontend/src/components/Util/TaskEditors/TaskEditor/task-diff-reducer.ts
```diff
@@ -74,100 +54,45 @@ function initializer([mainTask, subTasks]: [MainTask, readonly SubTask[]]): Stat
...
-      const { subTasks, diff, ...restState } = state;
-      const { newSubTask } = action;
-      const id = getNewSubTaskId();
-      const newSubTasks = [...subTasks, { id, ...newSubTask }];
-      const newDiff = {
-        ...diff,
-        subTaskCreations: diff.subTaskCreations.set(id, newSubTask),
-      };
-      return { ...restState, subTasks: newSubTasks, diff: newDiff };
-    }
```
SamChou19815, line 92: here is my hypothesis on what might go wrong.
you can see in the reducer that we create a new state based on the previous state, however, in your code, you update the state NOT based on the old state, but based on something in the prop.
This is potentially very problematic, because setState is async, and react may choose to batch update. When react batch updates, then any code that doesn't update based on previous state (aka. not using update function but directly give a new value) can mess up.

to solve the problem, you think you need to give up the approach that deep compares subtask content. Instead, you should compare using order, since we know for sure order is unique and it's immutable!

frontend/src/components/Util/TaskEditors/FloatingTaskEditor.tsx
ptwu: the main `taskData` here doesn't have a `children` property, so I think we're adding it here so it can match the `TaskMainData` props expected
image

SamChou19815 (approved):

PR #640, opened 2020-11-07
Merge pending invites into samwise-groups

pihucode:
### Summary

This pull request merges `samwise-group-pending-invites` into `samwise-groups`. Two string arrays are used to store invitation information: `invitees` holds emails of invited users, `inviterNames` holds the names of people who sent the invitation. These two arrays are parallel arrays such that `inviterNames[0]` is the name of the person who sent an invite for `invitees[0]`. 

Group invitations have the following behaviours:
- Inviting the same user multiple times will not result in multiple entries written to `invitees` or `inviterNames`. 
- A user cannot accept an invitation if they are already part of the group. 
- If the invited user rejects the invite, then the invitation data will be removed from both `invitees` and `inviterNames`. 
- If the invited user accepts the invite, then they will be added to the members list of the group and their invitation data will be removed from both `invitees` and `inviterNames`.

### Test Plan
Create invitations on the frontend dev server through the 'Add member' button in group view and check if data is correctly written to samwise-groups on Firestore.

### Breaking Changes
There is no longer a need for the `samwise-group-pending-invites` collection. Documents in `samwise-groups` now have `invitees` and `inviterNames` fields.
Firestore rules are updated so that only group members or invitees can read, write, and delete.


SamChou19815 (approved):
tested live on zoom, looks good

PR #641, opened 2020-11-07
Use default credentials for firebase functions in prod

SamChou19815:
### Summary 

Just remembered that we don't actually need to store credential in secrets. In the DTI hackathon project, I directly used:

https://github.com/AkashAryal/Flower/blob/98d233e60cba8be653722e99d7a5cb739b43db52/server/src/index.ts#L1-L9

and everything seems to work fine. I still leave the credential reading code so that we can still run stuff locally if we want to.

### Test Plan 

notest

### DIFF
--- a/functions/src/db.ts
+++ b/functions/src/db.ts
@@ -1,15 +1,17 @@
 import * as admin from 'firebase-admin';
+import { readFileSync } from 'fs';
 import Database from 'common/firebase/database';
-import serviceAccount from './firebase-adminsdk.json';
 
 if (process.env.TEST_MODE) {
   // eslint-disable-next-line no-console
   console.log('We are in test mode. Skip initializing firebase-admin.');
-} else {
+} else if (process.env.DEV) {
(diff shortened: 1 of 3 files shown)


functions/src/db.ts
```diff
@@ -1,15 +1,17 @@
...
 
 if (process.env.TEST_MODE) {
   // eslint-disable-next-line no-console
   console.log('We are in test mode. Skip initializing firebase-admin.');
-} else {
+} else if (process.env.DEV) {
   admin.initializeApp({
-    credential: admin.credential.cert(serviceAccount as admin.ServiceAccount),
+    credential: JSON.parse(readFileSync('./firebase-adminsdk.json').toString()),
     databaseURL: 'https://samwise-dev.firebaseio.com',
```
SamChou19815, line 11: The existing code is actually very problematic. For example, this line is probably the reason why some functions are not properly working in prod instance.

meganyin13 (approved):

PR #636, opened 2020-11-07
Order tasks in group/focus/future view

IshikaJain:
### Summary 

Ordered tasks in group and focus view where group tasks come first then all individual tasks ordered by id. 

### Test Plan 

`yarn test`. The added test tests the order comparator function is correct.

PR #645, opened 2020-11-09
SubTask -> Task migration release

SamChou19815:
### Summary 

It contains @ptwu's change on the subtask -> task frontend migration and a script to migrate existing DB data. The script is already run.

I also merged in the order change in #636, forgetting this release PR still exists... This PR sorts the tasks in the order of [group, repeating, one-time], then by order. Please make sure it doesn't break as well.

### Test Plan 

Please help test on staging site to ensure nothing breaks, including but not limited to:

- creating tasks
- editing subtask from focus view/future view
- deleting individual subtask in focus view/future view
- focus/complete individual subtask or entire task in focus/future view

### DIFF
--- a/common/firebase/database.ts
+++ b/common/firebase/database.ts
@@ -150,6 +150,8 @@ export default class Database {
 
   tasksCollection = (): Collection => this.db().collection('samwise-tasks');
 
+  subTasksCollection = (): Collection => this.db().collection('samwise-subtasks');
+
   groupsCollection = (): Collection => this.db().collection('samwise-groups');
 
   usersCollection = (): Collection => this.db().collection('samwise-users');
--- a/common/firebase/database.ts
+++ b/common/firebase/database.ts
@@ -150,8 +150,6 @@ export default class Database {
 
   tasksCollection = (): Collection => this.db().collection('samwise-tasks');
 
-  subTasksCollection = (): Collection => this.db().collection('samwise-subtasks');
-
   groupsCollection = (): Collection => this.db().collection('samwise-groups');
 
   usersCollection = (): Collection => this.db().collection('samwise-users');
--- a/common/types/action-types.ts
+++ b/common/types/action-types.ts
@@ -1,5 +1,5 @@
 import { Map } from 'immutable';
-import { Course, Settings, SubTask, Task, Tag, BannerMessageStatus, Group } from './store-types';
+import { Course, Settings, Task, Tag, BannerMessageStatus, Group } from './store-types';
 
 export type PatchTags = {
   readonly type: 'PATCH_TAGS';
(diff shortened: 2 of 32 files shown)


ptwu (changes requested):
clear focus is broken right now, giving me this error.

image

Seems to be caused by this function in `actions.ts`. Unsure why the `commit()` is getting called early.

https://github.com/cornell-dti/samwise/blob/d8b344c3b6291026c457d6711e91941cf96a6ee2/frontend/src/firebase/actions.ts#L480

ptwu (approved):
lgtm

meganyin13 (commented):
Really great work master hacker Peter but we should check repeating tasks. I noticed:
1. completing/deleting a subtask of repeating task automatically applies that for all tasks of that template. I think we should be forking it?
2. Pinning it results in the error bear page :(

Wow I also now can't get back in my samwise-dev account cuz it's stuck on error bear page lol I think some data is corrupted.

SamChou19815 (commented):
oof, looks like we have some tasks that only have children???

Screen Shot 2020-11-08 at 23 31 06

octocat-astonished

ptwu:
> oof, looks like we have some tasks that only have children???
> 
> Screen Shot 2020-11-08 at 23 31 06
> 
> octocat-astonished

Yikes. did you run the script on prod already?

SamChou19815:
> > oof, looks like we have some tasks that only have children???
> > Screen Shot 2020-11-08 at 23 31 06
> > octocat-astonished
> 
> Yikes. did you run the script on prod already?

no, it's from staging

SamChou19815:
@ptwu looks like this is the code to blame for broken repeating tasks:

https://github.com/cornell-dti/samwise/blob/86ca1f254f15e17063ddd169bacc95281acfdb86/frontend/src/components/TaskView/FutureView/FutureViewSubTask.tsx#L22-L47

you should not assume it's one time task

ptwu:
> > > oof, looks like we have some tasks that only have children???
> > 
> > 
> > Yikes. did you run the script on prod already?
> 
> no, it's from staging

Maybe I need to look at the script one more time. It might actually only be updating the children to the new format and not moving all the rest of the data.

meganyin13 (commented):
I am seeing that for group tasks, subtasks aren't showing on the group view task card (although it does show up on personal view). This is not push-blocking though.

ptwu:
PR should be ready for review now.

ptwu (approved):
Made sure to delete old tasks and played with new ones: both one-time and repeating. Things seem to be working as expected.

meganyin13 (approved):
looks very good! I've tested one time and repeating tasks and deleting/pinning/completing them and forking. Things seem to work. Also the group tasks subtask issue I mentioned earlier is also master so definitely not push blocking. Great job master hacker Peter! 
shipit_parrot

Ordering also looks good. I think we should disable dragging on group tasks because right now I can drag group tasks and one time tasks into group tasks section (although it doesn't actually get dropped there.)

jason-tung (commented):
samwisebug
can't change task completion from future view

ptwu (approved):

jason-tung:
samwisebug2

quickly clicking breaks marking repeating tasks as complete

ptwu:
> samwisebug2
> 
> quickly clicking breaks marking repeating tasks as complete

I was able to reproduce this bug in prod. I think we can separately make a PR to fix this in a different release.

jason-tung (approved):

meganyin13 (approved):
Amazing work master hacker mvp dev peter!! 
shipit_parrot

PR #646, opened 2020-11-09
Update clearFocus to await on database operations to complete

ptwu:
### Summary 

This pull request fixes the clear focus functionality caused by refactoring in #635. Previously, we did not await on the asynchronous operations of iterating through the tasks and subtasks needed to be cleared from focus, and `batch.commit()` would execute while we were still adding operations to the batch. This PR ensures that these asynchronous operations are completed before committing using `Promise.all`.

- [x] fixes Clear Focus functionality
- [x] fixes push-blocking issue in #645

### Test Plan 
s

Make some tasks and focus them. Complete them and see confetti. Click clear focus.

### DIFF
--- a/frontend/src/components/TaskView/FocusView/ClearFocus.tsx
+++ b/frontend/src/components/TaskView/FocusView/ClearFocus.tsx
@@ -34,7 +34,7 @@ function ClearFocus({ tasks }: Props): ReactElement | null {
   if (taskIds.length === 0) {
     return null;
   }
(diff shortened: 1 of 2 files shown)


SamChou19815 (approved):
I'm going to approve for now.
But you can see this entire `clearFocus` function was written back in the old days when tasks and subtasks are in separate collection. I think refactoring the codebase to make it more efficient in the new setup could be your task for this week.

PR #649, opened 2020-11-10
Send email notification

jason-tung:
### Summary 

Clicking heart or bell sends an email to the associated person from the person list. The email at the moment is very simple:
```
subject: `You got a ${variant}!`,
text: `${senderName} from ${group.name} is sending you a ${variant}!`,
```
We will need to place a Sendgrid key in `functions/src/sendgrid-key.json` and change the email in [`functions/src/send-email.ts`](https://github.com/cornell-dti/samwise/blob/e25835094e61b973ec14b765aedb87cb6cbdf460/functions/src/send-email.ts#L46). 

### Test Plan 

Playtested sending email to myself and peter.
Fiddle with it on staging, I suppose.


jason-tung:
Linter warnings are because there is no API key since it's gitignored

frontend/package.json
```diff
@@ -15,6 +15,7 @@
     "@fortawesome/free-solid-svg-icons": "^5.12.0",
     "@fortawesome/react-fontawesome": "^0.1.8",
     "common": "0.1.0",
+    "cors": "^2.8.5",
```
SamChou19815, line 18: We shouldn't need this

frontend/src/components/GroupView/MiddleBar/GroupViewMiddleBarMemberRow.tsx
```diff
@@ -1,16 +1,31 @@
 import React, { ReactElement } from 'react';
 import { faPlus } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import firebase from 'firebase';
```
SamChou19815, line 4: Import from firebase/app instead

functions/.gitignore
```diff
@@ -0,0 +1 @@
+sendgrid.env
```
SamChou19815, line 1: Not sure why you need this. Looks like you already gitignored the token json

functions/src/sendgrid-key.json
```diff
@@ -0,0 +1 @@
+{"SENDGRID_API_KEY":""}
```
SamChou19815, line 1: this should not be here

SamChou19815 (commented):
@jason-tung could you send me an email using the feature you built, so that I know it works?

functions/src/sendgrid-key.json
jason-tung: Is my CI code incorrect? I removed the file and the CI command to write a dummy file is still there but the CI is failing

SamChou19815: @jason-tung you need to add the CI command to all the jobs that might need the key. It's left as an exercise for the reader to figure out what are those CI jobs 😉

jason-tung: 🤔

SamChou19815 (approved):

PR #651, opened 2020-11-13
Fix FloatingTaskEditor checkbox and focus

ptwu:
### Summary 

This pull request fixes task completion in future view's FloatingTaskEditor. Previously, we could check the checkbox to complete the task once, but it would only update once per hover+click, and was not very responsive. This is because during #635, we added the `active` prop to the `TaskEditor` when using the FloatingTaskEditor under the assumption that we needed to in order to prevent race conditions in updating the state of the task in the reducer. However, this turned out to not be the root cause of the deleting / corrupted data problem, as the real issue was that we were not editing tasks in a principled way through the reducer. Therefore, this `active` prop can be removed, and solves the issue in one line. 😎

- [x] fixes issue pointed out by @jason-tung in #645.

### Test Plan 

XuLX5pIull


meganyin13 (approved):
:dev-peter:

PR #655, opened 2020-11-17
Avoid using position: absolute to construct TaskCreator. Close #618

### LINKED ISSUE #618: [code-quality] TaskCreator css code is terrible

SamChou19815:
### Summary 

After #617, #628, #630, #644, we are finally in a good position to fix the bad code in `TaskCreator`.

As a recap, the existing code is only `position: absolute` to hack code positions of varies components of `TaskCreator`, merely to overcome the difficulty to place some content before the `input` without input appearing after those content in JSX.

This PR almost completely rewrites the css related to positioning in `TaskCreator`. I also moved some jsx code around to make the layout make more sense.

### Test Plan 

Old:
old

New personal:
new-personal

New group:
new-group

### DIFF
--- a/frontend/src/components/TaskCreator/Picker.module.scss
+++ b/frontend/src/components/TaskCreator/Picker.module.scss
@@ -6,7 +6,6 @@
   position: relative;
   color: white;
   margin-left: 5px;
-  transform: translateY(-30px);
   z-index: 2;
 }
 
--- a/frontend/src/components/TaskCreator/__snapshots__/index.test.tsx.snap
+++ b/frontend/src/components/TaskCreator/__snapshots__/index.test.tsx.snap
@@ -5,7 +5,7 @@ exports[`TaskCreator matches snapshot 1`] = `
   className="TaskCreator TaskCreatorClosed"
 >
   <form
-    className="NewTaskWrap"
+    className="NewTaskWrap "
     onFocus={[Function]}
     onSubmit={[Function]}
   >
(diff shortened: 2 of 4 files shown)


meganyin13 (commented):
Something is a little off here... the cursor for the main task text & assign to search bar doesn't start at the left most side of the box (as it does on prod):
main-task-text
assign-to-searchbar

Also, the whole TaskCreator seems to have shifted down which is fine for group view but personal view is shifted a little too low 😕

SamChou19815:
> Something is a little off here... the cursor for the main task text & assign to search bar doesn't start at the left most side of the box (as it does on prod):
> 
> main-task-text
> 
> assign-to-searchbar
> 
> Also, the whole TaskCreator seems to have shifted down which is fine for group view but personal view is shifted a little too low 😕

Good catch. I will fix these when I have time. To all the future reviewers, this is a tricky refactor and needs more careful reviews!

meganyin13 (commented):
When comparing prod and staging side by side things are still a little bit shifted which is fine I think
prod-staging side by side

But the class tags are now a little too wide:
this PR
this PR
Prod
prod

Also this PR adds some space between `Add Subtasks (optional)` text and the below subtext
line break

meganyin13 (commented):
I think the `add class` and `add date` bubbles are shifted a little too much to the right?
this PR:
image
prod:
image

Also, I think there's actually less empty space at the bottom of TaskCreator which I think is better design 😄

meganyin13 (approved):
LGTM 🚀

PR #657, opened 2020-11-19
Rewrite the course-fetcher to make it completely in TS

### LINKED ISSUE #552: [maintenence] Replace Sam's script to fetch courses from course roster with TS code

SamChou19815:
## Summary 

Adapting the code from course plan: https://github.com/cornell-dti/course-plan/blob/master/src/requirements/fetcher.ts, which is also written by me.
No need to clone the fetcher repo and compile it. Everything is much faster.

Close #552.

## Test Plan 

New fetched json:
Screen Shot 2020-11-19 at 12 25 50


meganyin13 (approved):
Interesting statistics 📈

PR #664, opened 2020-11-21
sending email notif about task

jason-tung:
### Summary 

Clicking the bell icon now sends an email to the owner of the task as follows:
`${senderName} from ${group.name} is sending you a reminder to do ${task.name}!`

### Test Plan 
Playtest. Playtested w/ myself and one other.

### DIFF
--- a/frontend/src/components/GroupView/MiddleBar/GroupViewMiddleBarTaskQueue.tsx
+++ b/frontend/src/components/GroupView/MiddleBar/GroupViewMiddleBarTaskQueue.tsx
@@ -18,7 +18,7 @@ const EmptyTaskQueue = (): React.ReactElement => (
 
 function renderTaskList(tasks: readonly Task[]): ReactNode {
   return [...tasks].sort(sortTask).map((task) => {
-    return <GroupTask key={task.id} original={task} memberName="" />;
+    return <GroupTask key={task.id} original={task} memberName="" groupID="" memberEmail="" />;
   });
 }
 
--- a/frontend/src/components/GroupView/RightView/GroupTask.tsx
+++ b/frontend/src/components/GroupView/RightView/GroupTask.tsx
@@ -6,9 +6,11 @@ import styles from './GroupTask.module.scss';
 type Props = {
   readonly original: Task;
   readonly memberName: string;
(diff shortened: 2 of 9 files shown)


frontend/src/components/GroupView/MiddleBar/GroupViewMiddleBarTaskQueue.tsx
```diff
@@ -18,7 +18,7 @@ const EmptyTaskQueue = (): React.ReactElement => (
 
 function renderTaskList(tasks: readonly Task[]): ReactNode {
   return [...tasks].sort(sortTask).map((task) => {
-    return <GroupTask key={task.id} original={task} memberName="" />;
+    return <GroupTask key={task.id} original={task} memberName="" groupID="" memberEmail="" />;
```
SamChou19815, line 21: this doesn't look right...

jason-tung: this is this 
image
right? What is the issue?

SamChou19815: I don't think passing these dummy data is a good idea. What's your reason of doing this?

jason-tung: I copied the format from the memberName prop

SamChou19815: hmm the original code also doesn't have good practice... If you can explain to me why this is safe I can let this go through, but there needs to be a follow up task to clean this up.

jason-tung: Well, this belongs to the personal group task queue and it doesn't make sense to notify yourself about your own tasks, hence there is no bell icon on these tasks-- i.e. it doesn't necessarily need either of these fields. Although, I do also think that it is weird to use a GroupTask here since such is the case.

SamChou19815 (approved):

PR #670, opened 2020-12-03
Use DTI email as sender for group task reminder emails

SamChou19815:
### Summary 

It looks more professional, and emails are sent by sendgrid anyways.

### Test Plan 

👀


jason-tung (commented):
Need to change the SendGrid key because it still sends from my email on my end
image

jason-tung (approved):

PR #673, opened 2020-12-05
Ensure the group task progress is always at the bottom of the screen

SamChou19815:
### Summary 

Right now in the code, we are using RightView class twice, which is incorrect and caused we to have wrong layout with regard to replacing task progress at the bottom of the screen. This PR fixes that.

### Test Plan 

Screen Shot 2020-12-05 at 15 41 55


ptwu (approved):
LGTM

PR #676, opened 2020-12-05
Group view task scrolling on overflow

ptwu:
### Summary 

Previously, when rendering Group Tasks in Group View, we naively showed an entire task as if it were in focus view. However, subtask lists can get long, and we need to handle this case. 

This PR implements Group Task scrolling when there are too many subtasks for a given task (The Figma design shows 2 subtasks maximum, so we render the original styles when there are 2 subtasks or fewer, and start a scrollbar overflow otherwise). This was actually a nontrivial task, since simply adding an `overflow-y: scroll` attribute to the Group Task actually caused it to exhibit unexpected, weird behavior, [as seen in this Stack Overflow link](https://stackoverflow.com/questions/6421966/css-overflow-x-visible-and-overflow-y-hidden-causing-scrollbar-issue). As such, we needed to circumvent this by adding a parent div to each group task, which has its own max height attribute and controls for overflow. After this change was introduced, we also needed to fix the Task Queue rendering as the multiple scrollbars led to an extremely buggy UI on Windows. To do this, we simply pass down a prop for whether a group task is in group view and use the original styles if not. 

- [x] implements overflow scrolling for group tasks.

### Test Plan 

dxkEJTbFze

Go to GroupView and create a task with a lot of subtasks. See that a scroll bar appears for those tasks.

### DIFF
--- a/frontend/src/components/GroupView/index.tsx
+++ b/frontend/src/components/GroupView/index.tsx
@@ -1,11 +1,5 @@
 import React, { ReactElement, useState, useEffect } from 'react';
-import type {
-  Group,
-  SamwiseUserProfile,
-  Task,
-  GroupTaskMetadata,
-  SubTask,
-} from 'common/types/store-types';
+import type { Group, SamwiseUserProfile, Task, GroupTaskMetadata } from 'common/types/store-types';
 import type { FirestoreUserData } from 'common/types/firestore-types';
 import MiddleBar from './MiddleBar';
 import RightView from './RightView';
@@ -73,7 +67,6 @@ const GroupView = ({ group, changeView }: Props): ReactElement => {
                 date: docData.date.toDate(),
                 group: docData.group,
               } as GroupTaskMetadata,
-              children: [] as readonly SubTask[],
             } as Task;
           })
         );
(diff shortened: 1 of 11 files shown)


frontend/src/components/GroupView/index.tsx
```diff
@@ -1,11 +1,5 @@
...
-  Group,
-  SamwiseUserProfile,
-  Task,
-  GroupTaskMetadata,
-  SubTask,
-} from 'common/types/store-types';
+import type { Group, SamwiseUserProfile, Task, GroupTaskMetadata } from 'common/types/store-types';
 import type { FirestoreUserData } from 'common/types/firestore-types';
 import MiddleBar from './MiddleBar';
 import RightView from './RightView';
```
SamChou19815, line 5: merge in master into this branch so that these changes from #672 won't reappear here.

SamChou19815:
so far it seems to work fine for me on mac, but now it seems to cause the group tasks to have different widthx

ptwu:
this PR should be ready for review now.

SamChou19815:
right now tasks in group view are having different heights, which is bad.

SamChou19815 (approved):
Looks good

PR #677, opened 2020-12-05
Groups overflow

### LINKED ISSUE #585: [group-task] Making many groups overflows the SideBar
**Describe the bug**
If you make a lot of groups, each group icon gets rendered below the previous one, and causes the New Group button to move down. Making many of them may cause the New Group button to become inaccessible.

**To Reproduce**
Keep making groups (you can have more than one group for a class). On my screen, when I have more than 5, the New Group button clips the settings button.

**Screenshots**
image

jason-tung:
### Summary 

Fix bug where creating too many groups resulted in an ugly overflow

fixes #585 

### Test Plan 

👀👀

example


jason-tung:
Was going to make the Add icon scroll with the groups do something similar to how `FloatingTaskEditor` to find out where to place the dropdown but decided against that because there would be more complicated logic to tell whether or not the dropdown should be visible at all. e.g. what if the scrollbar is all the way at the top instead of at the bottom?

frontend/src/components/GroupView/groupview-sizes.scss
```diff
@@ -5,6 +5,11 @@ $progressbar-height: 80px;
 $sidebar-width: 80px;
 $sidebar-header-height: 24px;
 $sidebar-header-margin: 14px;
+$sidebar-personal-button-height: 15vh;
+$sidebar-group-manager-height: 85vh;
```
SamChou19815, line 9: why personal view button doesn't have a fixed height

frontend/src/components/GroupView/groupview-sizes.scss
```diff
@@ -5,6 +5,11 @@ $progressbar-height: 80px;
...
+$add-button-height: 80px;
+$expanding-div-base-height: 75px;
+$links-height: 62px;
```
SamChou19815, line 12: can we make these all divisible by 4. Random numbers don't look good

frontend/src/components/SideBar/index.module.scss
```diff
@@ -67,7 +74,40 @@
...
+  flex-basis: $expanding-div-base-height;
+  flex-shrink: 0;
+}
+
+.GroupIcons {
+  width: $sidebar-width;
+  overflow-y: auto;
+  max-height: calc(
+    #{$sidebar-group-manager-height} - #{$expanding-div-base-height} - #{$links-height} - #{$add-button-height} -
+      20px
```
SamChou19815, line 88: what's this final 20px?

SamChou19815 (commented):
some nits

frontend/src/components/GroupView/groupview-sizes.scss
jason-tung: What's the thinking behind making it divisible by 4? Just convention?

SamChou19815: some mature design system like material design has this systematic guideline: https://material.io/design/layout/spacing-methods.html#spacing

Having some random number like 62 is usually a sign that you tweaked the pixel value until it looks somehow correct, but without any systematic principle behind it. If we have those unprincipled number all over the place, it would be impossible fo enforce design consistency.

SamChou19815 (approved):
Other than some random css pixel values, the PR looks good to me. Please try to fix them before merging.

frontend/src/components/SideBar/index.module.scss
jason-tung: the height of the new group span

PR #678, opened 2020-12-06
Change course json file to make it include no semester name

SamChou19815:
### Summary 

When samwise is put into maintenance mode, we want updates to be as smooth as possible. Right now updating course json is already a headache because cornell keeps changing format every semester.

However, there is one thing we can easily do: remove the semester prefix of the json file, so that during maintenance, we have one less thing to change.

### Test Plan 

CI


ptwu (approved):
😭😭

PR #680, opened 2020-12-08
Fix Focus View Filtered Subtasks SEV

### LINKED ISSUE #679: Subtasks Randomly Disappear
**Describe the bug**
Performing common operations will cause subtasks to disappear.

**To Reproduce**
Create a new task due tomorrow. Do not pin the main task to focus.
Create two subtasks for the task. Pin only the first subtask to focus; do not pin the second.
Check off the first subtask.
Observe that the second subtask will disappear.

**Environment (optional)**
Production
Windows 10, Google Chrome (latest updates on both)

ptwu:
### Background (What happened)

Largely resulting from the changes in #635 (migration of SubTask -> Task in the types), checking off a pinned subtask in a task in focus view can delete other subtasks unintentionally. This is because since subtasks are now a part of the `Task` type, the filter between completed and uncompleted tasks in focus view actually creates two different forks of the same task. This is bad, since we apply changes through the `task-diff-reducer` using its hook which needs to have `Task` data. Also, as pointed out by Sam, since I nuked the subtask diffing functions, the task will not handle these forks correctly, and we now have unexpected behavior.

This issue was pointed out by Michael @mt-xing in #679 (with screenshots on how to reproduce on `master` atm), and this PR aims to fix it.

### Summary 

In order to implement a fix to this problem, we pass the whole `Task` as `Task<TaskMetadata>` into the `TaskEditor` as an optional argument if it is a filtered task from focus view. This way, a filtered task can access its parent's data no matter what. Some undefined checks are implemented to deal with this, which may cause a bit of technical debt, but comments are provided so future developers can understand what the prop functions as. We then use the reducer to get the task data for the filtered part as well as the whole task so that it remains the single source of truth, and since the `OneSubTaskEditor` only cares about the subtask data, we can pass the callbacks from the whole task's reducer hook but sneakily render a list of these subtasks using the filtered data. This way, we can resolve the overwriting problem, as it behaves as if it is in FutureView again.

- [x] fixes #679

### Test Plan 

Create a task with several subtasks. Pin them to focus, and start checking them off. See that there is no unexpected deletion anymore. 

demo

### Notes 

RFC: An unintentional side effect (as of right now) of this implementation is that now the `OneSubTaskEditor` checkboxes seem hyper-sensitive to whether the task editor is currently active, i.e. there is a mouse pointer inside of it. This can be seen in the demo GIF, as I have to take my cursor outside of the editor card in order to see the changes happen. This does not affect functionality, however.

### DIFF
--- a/frontend/src/components/Util/TaskEditors/TaskEditor/index.tsx
+++ b/frontend/src/components/Util/TaskEditors/TaskEditor/index.tsx
@@ -17,6 +17,10 @@ import {
 import { NONE_TAG } from 'common/util/tag-util';
 import { ignore } from 'common/util/general-util';
 import { getTodayAtZeroAM, getDateWithDateString } from 'common/util/datetime-util';
+import {
+  getFilteredCompletedInFocusTask,
+  getFilteredNotCompletedInFocusTask,
+} from 'common/util/task-util';
 import OverdueAlert from '../../../UI/OverdueAlert';
 import {
   confirmRepeatedTaskEditMaster,
@@ -62,6 +66,7 @@ type OwnProps = DefaultProps & {
   readonly memberEmail?: string; // only supplied if task is a group task
   readonly groupID?: string; // only supplied if task is a group task
   readonly wholeTaskData?: Task<TaskMetadata>; // only supplied if task is a focus task
+  readonly isFocusTaskAndCompleted?: boolean; // only supplied if task is in focus view
 };
 type Props = OwnProps & {
   // subscribed from redux store.
@@ -98,6 +103,7 @@ function TaskEditor({
   memberEmail,
   groupID,
   wholeTaskData,
+  isFocusTaskAndCompleted,
 }: Props): ReactElement {
   const { onChange, removeTask, onSaveClicked } = actions;
   const initTaskDate = initTaskData.date;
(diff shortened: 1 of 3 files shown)


SamChou19815 (changes requested):
subtask text editing seems to be broken by this diff. Can't figure out the reason yet.

frontend/src/components/Util/TaskEditors/TaskEditor/index.tsx
```diff
@@ -88,9 +97,18 @@ function TaskEditor({
...
+    wholeTaskData ? ({ ...wholeTaskData, date: initTaskDate } as TaskMainData) : initTaskData,
+    wholeTaskData ? false : active ?? false,
+    wholeTaskData ? ignore : onChange ?? ignore
+  );
+
+  // only for use if the current task is a filtered completed/uncompleted task in focus view
+  const filteredTaskDiffReducerData = useTaskDiffReducer(
     initTaskData,
     active ?? false,
     onChange ?? ignore
```
SamChou19815, line 114: might know the reason why editing is broken now:

rendering uses filteredTaskDiffReducerData, but all the updates are applied with the functions returned by the first `filteredTaskDiffReducerData `

ptwu:
> subtask text editing seems to be broken by this diff. Can't figure out the reason yet.

yikes. It seems to work in future view still, so it's probably caused by my use of two different calls to `useTaskDiffReducer`

ptwu:
Subtask editing seems to work again by my playtesting.

SamChou19815 (approved):
looks good

PR #683, opened 2020-12-11
Fix TaskCreator and member assignment bugs

pihucode:
### Summary
This pull request fixes `TaskCreator` bugs.

- [x] Fixed `TaskCreator` not closing when opened through the 'Assign new task' button.
- [x] Group tasks are now correctly assigned when using the 'Assign new task' button.
- [x] Manual assignment takes precedence over auto-filled assignment through the button.

### Test Plan
`TaskCreator` now closes correctly.
close_taskcreator

Manual member assignment takes precedence.
assignment_precedence


SamChou19815 (approved):
Seems to be a good short term solution. In the long term we might want this to be controlled by some global state so that we don't have two competing opened state.

PR #685, opened 2020-12-12
Convert task creator to functional component

SamChou19815:
### Summary 

I'm about to setup some hooks to allow better communications between task creator and its sibling components. This PR refactors TaskCreator to functional component to prepare for that change.

The conversion itself is straight forward. Most of the time it's just changing syntax, so it should be easy to review.

### Test Plan 

Test it in personal view and group view. Should both work.


meganyin13 (approved):
LGTM

PR #686, opened 2020-12-12
Make assign button (+) in the middle bar work

SamChou19815:
### Summary 

Currently, the right view in group view is communicating with TaskCreator by some ad hoc local state. Now if we want to add another similar communication for the purpose of implementing the + button in MiddleView, we need to duplicate the logic, which is bad.

It turns out that we can use React context to lift the state up, and use hooks to _reactively_ communicate with each other. This PR refactors the existing code to use hooks instead, and use the new setup to quickly implement the functionality for the assign button in the MiddleBar.

### Test Plan 

- Try opening and closing task creator both in personal and group view.
- Try to click the assign button in both MiddleBar and RightView

### DIFF
--- a/frontend/src/components/GroupView/MiddleBar/GroupViewMiddleBarMemberRow.tsx
+++ b/frontend/src/components/GroupView/MiddleBar/GroupViewMiddleBarMemberRow.tsx
@@ -5,6 +5,7 @@ import firebase from 'firebase/app';
 import SamwiseIcon from '../../UI/SamwiseIcon';
 import styles from './GroupViewMiddleBarMemberRow.module.scss';
 import ProfileImage from './ProfileImage';
+import { useTaskCreatorContextSetter } from '../../TaskCreator';
 
 type Props = {
   memberName: string;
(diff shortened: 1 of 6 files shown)


frontend/src/components/GroupView/MiddleBar/GroupViewMiddleBarMemberRow.tsx
```diff
@@ -26,6 +27,14 @@ const sendNotification = (variant: string, email: string, group: string): void =
...
 
 const Member = ({ memberName, profilePicURL, group, email }: Props): ReactElement => {
+  const setTaskCreatorContext = useTaskCreatorContextSetter();
+
+  const openTaskCreator = (): void =>
+    setTaskCreatorContext({
+      taskCreatorOpened: true,
+      assignedMembers: [{ email, name: memberName, photoURL: profilePicURL }],
+    });
+
```
SamChou19815, line 37: with the hooks and context setup, you only need these few lines!

meganyin13 (approved):

PR #688, opened 2020-12-12
Disallow group task completion if not the assignee

ptwu:
### Summary 

This PR implements group task completion behavior for group tasks, by disallowing task & subtask completion and undoing completion if you are not the assignee of the subtask. It computes whether this is possible in the `TaskEditor` component and compares the owner of the group task with the current app user email, and passes this down to the `MainTaskEditor` and `OneSubTaskEditor` respectively.

### Test Plan 

CVE9tVd5lE

Create a group and have some friends to be in the group with you. Make tasks for them and some subtasks. Try checking their subtasks (it shouldn't work). See that you can still check your own subtasks.


SamChou19815 (approved):
last launch blocker resolved 🎉

PR #691, opened 2020-12-12
Add flex-shrink to group task row to unbreak group task in safari

SamChou19815:
### Summary 

Credit: https://stackoverflow.com/questions/32689686/overlapping-css-flexbox-items-in-safari

### Test Plan 

Before:
Screen Shot 2020-12-12 at 15 20 22
After:
Screen Shot 2020-12-12 at 15 20 27


mt-xing (approved):
Lmao Safari.

I think if we use the `flex: ` shorthand this becomes less of a problem, because that shorthand also sets the shrink property, but since we don't set any flex properties on this element, it ended up happening

PR #694, opened 2020-12-12
Ban all group task tag edit

SamChou19815:
### Summary 

Pass the type down for a more reliable check.

Bug fixed:

- group task no longer grabbable
- group task tag no longer editable

### Test Plan 

👀


ptwu (approved):
Looks good, although I can still drag the group task in future view.

OthLyL2a3q

PR #696, opened 2020-12-12
Give each group task a tag by using classId to show color with best effort

SamChou19815:
### Summary 

Using class id as the tag, so that people can see some color as long as they have the class tag.

### Test Plan 

Try to create a group in a real class and see the color.

### Note

There is a bug in the context when the group info like group id and group code is not updating when you switch between groups, I will fix this in another diff


ptwu (approved):
Looks good, apart from completed tasks showing up in gray

image

PR #698, opened 2020-12-12
Fix button alignment

pihucode:
### Summary
Fix button alignment in TaskCreator and align icons in MemberRow of GroupView.

### Test Plan 

image
image


meganyin13 (approved):
looks good

PR #700, opened 2020-12-12
Fix mobile view to be usable after group tasks launch

ptwu:
### Summary 

In anticipation of our soon launch, we still want mobile to be usable. To do this, we just remove the group tasks feature on mobile devices :( since we have no mobile designs.

### Test Plan 

Before:
image

After:
image


meganyin13 (approved):
:dev-peter:

PR #707, opened 2020-12-17
Unbreak the left text in settings launched from sidebar

SamChou19815:
### Summary 

Right now the settings launched from sidebar is broken:
Screen Shot 2020-12-16 at 20 18 39
For reference, the settings launched from personal view is fine:
Screen Shot 2020-12-16 at 20 18 42

This PR fixes the problem that the left text is not shown. It is caused by some style being cascading down. This bug shows again that we should really avoid all those css blanketly targeting one type of html tag.

### Test Plan 

Shown now:
Screen Shot 2020-12-16 at 20 20 47

### DIFF
--- a/frontend/src/components/SideBar/AddGroupTags.tsx
+++ b/frontend/src/components/SideBar/AddGroupTags.tsx
@@ -1,6 +1,7 @@
 import React, { ReactElement } from 'react';
 import { Tag } from 'common/types/store-types';
 import { store } from '../../store/store';
+import sideBarStyles from './index.module.scss';
 import styles from './AddGroupTags.module.scss';
 import { createGroup } from '../../firebase/actions';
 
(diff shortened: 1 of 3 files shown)


ptwu (commented):
image

Still not working for me on Windows (Chrome) :/

SamChou19815:
> image
> 
> Still not working for me on Windows (Chrome) :/

the screenshot shows it's working. It's only making text like "add classes", "class tags" being visible again @ptwu

ptwu:
> > image
> > Still not working for me on Windows (Chrome) :/
> 
> the screenshot shows it's working. It's only making text like "add classes", "class tags" being visible again @ptwu

oops, I read too fast and thought the classic layout (with the same type of buttons) was what the PR intended to accomplish. LGTM

ptwu (approved):

PR #709, opened 2020-12-17
Eliminate all stylelint css warnings

SamChou19815:
### Summary 

Remove the last 2 `no-descending-specificity` warnings by replacing the implicit button styling with explicit classnames.

We now eliminated all `no-descending-specificity` warnings, so we can finally convert the warning into error!

### Test Plan 

the css is still being correctly applied:
Screen Shot 2020-12-16 at 20 45 12


meganyin13 (commented):
~~Oops I think this is a regression from #708~~ nvm seems to be gone now... I refreshed

image

meganyin13 (approved):

PR #710, opened 2020-12-17
Unbreak add group tags selector

SamChou19815:
### Summary 

Missed this button when doing #708.

### Test Plan 

Broken:
Screen Shot 2020-12-16 at 21 40 47
Fixed:
Screen Shot 2020-12-16 at 21 42 08


meganyin13 (approved):
yay for getting rid of `!important`!

PR #715, opened 2020-12-21
Refactor task related firestore actions into common package

SamChou19815:
### Summary 

Move the code there so that it can be used by both frontend and functions code. Most of the change is just moving around code.

### Test Plan 

Check creating, editing task, especially with forked ones, still works.


meganyin13 (approved):
LGTM 🚀

PR #719, opened 2020-12-26
Refactor all firestore actions into common actions

### LINKED ISSUE #534: [RFC] Abstract away firebase operations in frontend code
### Motivation

The `common` package contains some common primitive firebase code. It is designed to be the place where we put common firebase code like adding a task, editing a tag, etc. In this way, the same code can be used by
both frontend and firebase functions. 

The setup should also allow us to make interfaces for DB related ops, so that the firestore logic can be mocked away for deterministic testing. e.g.

```ts
interface SamwiseDatabaseService {
  addTag(id: string, name: string): Promise<void>;
  
  // ...
}

// For prod code
class SamwiseDatabaseServicePoweredByFirestore implements SamwiseDatabaseService {
  async addTag(id, name) {
    await firestore.collection(tag).doc(id).add({ name });
  }
  
  // ...
}

// For test
class SamwiseDatabaseServiceForTest implements SamwiseDatabaseService {
  readonly addedTags: [string, string][]

  addTag(id, name) { this.addedTags.push(id, name); }
}

it('can add tag', async () => {
  render(<TagAdder />);
  await typeSomeGarbageAndClickButtonToAddTag();
  expect(mockedDBService.addedTags[0]).toEqual(['3', 'tag 1']);
});
```

### Complications

Unfortunately, the reality is more complicated than this ideal picture. In `frontend` package, we use firestore instance provided by `firebase` package. In `functions` package, we use firestore instance provided by `firebase-admin` package. These two firestore instances' public APIs are exactly the same most of the times, but they still have some slight differences. As a result, their TS typedefs have slight differences, which makes it impossible to assign one to the other.

To solve this problem, I introduced a thin [interface layer](https://github.com/cornell-dti/samwise/blob/master/common/firebase/database.ts) that only keeps the common types between firestore from `firebase` and firestore from `firebase-admin`, then we can use TS's structural typing to make both firestore instance assigneable to this interface. The interface allows us to extract some common code, so we can finally share some code between frontend and functions. The layer was built in late FA19 fall, and it doesn't have any new improvements since then.

### Goals

1. All the firebase actions code in `frontend` no longer use the firestore from `firebase/app`. Instead, they should be run against a common interface type like `FirestoreCommon`.
2. All the firebase actions should no longer be separate functions. Instead, they should be put in a class that implements an interface like `SamwiseDatabaseActions`. It would allow us to decouple firestore logic and frontend logic, and allow us to mock the actions in tests.

### Benefits

- Deterministic testing
- Reuse complex actions code for `functions`, so that features that auto-focusing repeating tasks can be implemented without copy-pasting a lot of code between `frontend` and `functions`.

### Setup

- We define a React context `SamwiseDatabaseActionsContext` that takes in a class that implements the `SamwiseDatabaseActions` interface.
  - In prod code, we pass in `SamwiseDatabaseServicePoweredByFirestore`
  - In test code, we pass in `SamwiseDatabaseServiceForTest`
  - We should wrap the entire application code with this context, just like the `ReactReduxContext`
- In the application code that needs to call firebase code, we use hooks.
  - We define a hook `useSamwiseDatabaseActions()` that just reads the actions object in the context
  - Application code can do something like `const actions = useSamwiseDatabaseActions(); const handler = () => actions.deleteTag(id);`

SamChou19815:
### Summary 

Move the code there so that it can be used by both frontend and functions code. Most of the change is just moving around code.

Close #534.

### Test Plan 

Check the following features

- Clear focus
- Complete task in focus
- Reorder tasks in focus
- Complete onboarding
- Set canvas calendar
- Import course exam

### DIFF
--- a/common/firebase/common-actions.ts
+++ b/common/firebase/common-actions.ts
@@ -1,11 +1,16 @@
+import { Map } from 'immutable';
 import {
+  FirestoreCommonTask,
   FirestoreGroup,
   FirestoreTag,
   FirestoreTask,
   FirestoreUserData,
 } from '../types/firestore-types';
 import {
+  BannerMessageIds,
+  Course,
   Group,
+  OneTimeTaskMetadata,
   PartialTaskMainData,
   RepeatingTaskMetadata,
   State,
@@ -18,6 +23,9 @@ import Database, { WriteBatch } from './database';
 import OrderManager from './order-manager';
 import { error, ignore } from '../util/general-util';
 import { NONE_TAG_ID } from '../util/tag-util';
+import { subTasksEqual } from '../util/task-util';
+import { Action } from '../types/action-types';
+import { patchTags, patchTasks } from '../store/actions';
 
 type WithoutIdOrder<Props> = Pick<Props, Exclude<keyof Props, 'id' | 'order'>>;
 type WithoutId<Props> = Pick<Props, Exclude<keyof Props, 'id'>>;
@@ -57,6 +65,11 @@ export default class Actions {
     const order = await this.orderManager.allocateNewOrder('tasks');
     return { ...source, owner, order };
   };
+
+  private mergeWithOwner = <T>(obj: T): T & { readonly owner: string } => ({
+    owner: this.getUserEmail(),
+    ...obj,
+  });
 
   /*
    * --------------------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


ptwu (commented):
image

can't seem to log in... getting this error in console.

SamChou19815:
@ptwu try this in local and see where it throws exactly. It looks like some data validation has failed, which shouldn't be caused by this PR.

meganyin13 (approved):
Approving, but I have discovered things are broken but if they are, they are also broken on prod.
> Check the following features
> 
> * Clear focus

Clearing focus seems to remove the task from the UI entirely. Also exists on master.

https://user-images.githubusercontent.com/20008134/103161265-1c9e9000-47ad-11eb-9418-56c5187fba8f.mov

> * Complete task in focus

Seems to work.

> * Reorder tasks in focus
Somehow tasks are no longer being reordered. Also exists on master

https://user-images.githubusercontent.com/20008134/103161270-33dd7d80-47ad-11eb-9647-309d8e55bbb6.mov

> * Complete onboarding

Seems to work.

> * Set canvas calendar
> * Import course exam

I've tried these but school is not in session and there were no obvious errors so I will assume it works...

PR #720, opened 2020-12-27
Fix clear focus

meganyin13:
### Summary

Currently, when you click clear focus on a set of completed tasks, it actually appears to remove those tasks from the UI. By inspection, it's because these lines should be an `update` rather than a `set` since `set` will set the whole document to just those children.

https://user-images.githubusercontent.com/20008134/103161338-4c01cc80-47ae-11eb-8fb7-5d7565586334.mov

### Test Plan

Create some tasks, complete them,

 and then click clear focus.

### Notes
I'll move this into `common/firebase/common-actions.ts` once #719 is merged.


SamChou19815 (approved):
LGTM

PR #721, opened 2020-12-27
Use clsx to concat (sometimes conditional) css classnames

SamChou19815:
### Summary 

Using template literals can concat css well sometimes, but when there are conditional css classnames, it gets messy very quickly. This PR replaces those template literal css classnames with the `clsx` library, which can handle the conditional css very well.

### Test Plan 

Check the affected components' css is still fine. Read the changes to convince yourself that it's correct.


meganyin13 (approved):
LGTM
samwise-css-error

Issue #333, opened 2019-10-23
Overdue Repeating Tasks Don't Get Focused


mt-xing:
**Describe the bug**
Normally, when the due date of a task arrives, it is automatically focused. Thus, when the task becomes overdue, the task will be in your focus with the red exclamation mark.
However, when the due date of a non-forked instance of a repeating task arrives, nothing happens. If the due date passes, then the instance just disappears off your feed entirely.

Blocks #313 

**To Reproduce**
Create a repeating task where there is an instance tomorrow. Come back in 2 days. Note that nothing is in focus and there is no indication you had a task due yesterday.

SamChou19815:
Blocked by #368

PR #722, opened 2020-12-27
Remove dependencence on store store in firestore common actions

SamChou19815:
### Summary 

These actions should be able to be called on functions, which doesn't have a redux store.

Therefore, I make the common actions to take full task as parameters instead of task id. Then in the frontend actions, I pass in the task fetched from redux store.

The methods refactored in this diff is important to implement #333.

### Test Plan 

Check task editing and forking still works.


meganyin13 (approved):
LGTM 🚀

PR #723, opened 2020-12-28
Auto-focus repeating task by firebase functions

### LINKED ISSUE #333: Overdue Repeating Tasks Don't Get Focused

SamChou19815:
### Summary 

Close #333, our last open issue.

Implemented via a firebase function that auto focuses the repeating task by forking it. It reuses the existing common util function to tell whether today is on the repeating tasks' repetition schedule.

### Test Plan 

I created a repeating task that repeats on today, and run the function manually. It indeed got focused.

### DIFF
--- a/functions/src/focus-today-task.ts
+++ b/functions/src/focus-today-task.ts
@@ -1,6 +1,10 @@
+import { FirestoreMasterTask } from 'common/types/firestore-types';
+import { RepeatingTaskMetadata, Task } from 'common/types/store-types';
+import Actions from 'common/firebase/common-actions';
+import { dateMatchRepeats } from 'common/util/task-util';
 import database from './db';
 
-const focusTasksThatAreDueToday = async (): Promise<void> => {
+const focusOneTimeTasksThatAreDueToday = async (): Promise<void> => {
   const todayAtZero = new Date();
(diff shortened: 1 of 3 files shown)


functions/src/focus-today-task.ts
```diff
@@ -18,4 +22,56 @@ const focusTasksThatAreDueToday = async (): Promise<void> => {
   batch.commit();
 };
 
+const focusRepeatingTasksThatAreDueToday = async (): Promise<void> => {
+  const todayAtZero = new Date();
+  todayAtZero.setHours(0, 0, 0, 0);
+  const querySnapsot = await database
```
meganyin13, line 28: I think you mean `querySnapshot`

meganyin13 (approved):

PR #741, opened 2021-03-27
RFC: Deploy to prod on merging to master

SamChou19815:
### Summary 

Samwise is now in maintenance mode. Updates are rare and they are all coming from course json updates. Therefore, it's a big burden to first merge diff into `master` and then into `release`.

I propose that merging to master deploys directly to prod, so I delete the release workflow in this PR and moves that into the `cd-workflow`. We can increase the required review for merging to master to 2. We already have a deploy preview, so it's very easy to see whether things work for reviewers.

### Test Plan 

👀


meganyin13 (approved):
Interesting case that when a project goes into production we may not need a staging environment anymore since we're not actively doing feature development. Only approving this since we keep approving reviews at 2 per PR and have the deploy preview site on every PR 👍

ptwu (approved):
seems sensible

